PR #22, opened 2018-07-06
Fix up serial number detection

antylama:
For disk like this: /dev/disk/by-id/ata-WDC_WD10JFCX-68N6GN0_WD-WX31A76R3KFS.


jaypipes:
++ thanks much! I'll try to get some unit tests created for the DiskSerialNumber() function so we can put some safeguards in place.

PR #41, opened 2018-08-09
Rework cobra CLI

retr0h:
* Moved the CLI into cmd package.
* Broke out each subcommand into own file.
* Created a version subcommand, which allows the binary to be built
  with version information when passed the following LDFLAGS.

    LDFLAGS := -s -w \
	-X 'main.version=$(VERSION)'
	-X 'main.buildHash=$(GITCOMMIT)'
	-X 'main.buildDate=$(BUILDDATE)'


retr0h:
Also added a sexy command banner lol

```
          __
 .-----. |  |--. .--.--.--.
 |  _  | |     | |  |  |  |
 |___  | |__|__| |________|
 |_____|

Discover hardware information.

https://github.com/jaypipes/ghw

Usage:
  ghwc [flags]
  ghwc [command]

Available Commands:
  block       Show block storage information for the host system
  cpu         Show CPU information for the host system
  gpu         Show graphics/GPU information for the host system
  help        Help about any command
  memory      Show memory information for the host system
  net         Show network information for the host system
  topology    Show topology information for the host system
  version     Display the version of gofile

Flags:
      --debug   Enable or disable debug mode
  -h, --help    help for ghwc

Use "ghwc [command] --help" for more information about a command.
```

cmd/root.go
```diff
@@ -0,0 +1,93 @@
...
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "ghwc",
+	Short: "ghwc - Discover hardware information.",
+	Long: `
+          __
+ .-----. |  |--. .--.--.--.
+ |  _  | |     | |  |  |  |
+ |___  | |__|__| |________|
+ |_____|
```
jaypipes: lol, this makes me think of someone giving someone else the middle finger.

cmd/version.go
```diff
@@ -0,0 +1,35 @@
...
+
+const debugHeader = `
+Date: %s
+Build: %s
+Version: %s
+Git Hash: %s
+`
+
+// versionCmd represents the version command
+var versionCmd = &cobra.Command{
```
jaypipes: Cool. I'd wanted to do something like this. Thanks for adding it! :)

vendor/vendor.json
```diff
@@ -21,5 +21,5 @@
 			"revisionTime": "2017-05-08T18:43:26Z"
 		}
 	],
-	"rootPath": "github.com/jaypipes/ghw/ghwc"
+	"rootPath": "github.com/jaypipes/ghw"
```
jaypipes: So, the reason I had the ghwc/vendor folder was that I did not want to make cobra a dependency for the ghw library itself. I wanted to have as few (or no) dependencies for the importable library as possible. Is there a way to indicate that the dependency is only for the ghwc example client tool and not the library itself?

cmd/root.go
retr0h: lolol

vendor/vendor.json
retr0h: Ah, I seee.  Now I understand what ghwc was for.

PR #44, opened 2018-08-20
Removed build tags

retr0h:
> Build tags and file suffixes overlap in fuctionality. For example, a
  file called mypkg_linux.go that contained the build tag // +build linux
  is redundant.

> In general, when choosing between a build tag or a file suffix, you
  should choose a file suffix when there is an exact match between the
  platform or architecture and the file you want to include.

  [1] https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

### DIFF
--- a/block_linux.go
+++ b/block_linux.go
@@ -1,5 +1,3 @@
-// +build linux
-//
 // Use and distribution licensed under the Apache license version 2.
 //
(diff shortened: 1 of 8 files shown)


jaypipes:
Thanks Dewbert :)

Issue #45, opened 2018-08-25
Correctly parse the mountpoint in /etc/mtab


thehowl:
Mountpoint with spaces, tabs, newlines exist. The solution to that is octal encoding:

>Therefore these characters are encoded in the files and the getmntent function takes care of the decoding while reading the entries back in. '\040' is used to encode a space character, '\011' to encode a tab character, '\012' to encode a newline character, and '\\\\' to encode a backslash.

From: https://www.gnu.org/software/libc/manual/html_node/mtab.html

I think it's generally safe to assume that `mtab`, being generated and not hand-written, will usually only contain these exceptions, although it's probably best to check if `getmntent` in libc parses the mountpoint generically (ie. by converting the octal to a byte and then using it as the proper byte) or it handles these cases only.

jaypipes:
Cool, thanks for letting me know, @thehowl :) I'll see what I can do to put some tests together and correct. Thanks, again!

jaypipes:
@thehowl Hi! Please let me know if PR #50 addresses this issue. Thanks!

PR #50, opened 2018-09-03
Handle newline, tab and space in mtab entry

jaypipes:
Adds specialized parsing for mountpoints in mtab entries that contain
spaces, tabs and newlines. These characters will be encoded as their
octal-to-string representations by the glibc, and we need to decode
these values accordingly.

Issue #45

### DIFF
--- a/block_linux.go
+++ b/block_linux.go
@@ -250,25 +250,45 @@ func PartitionInfo(part string) (string, string, bool) {
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
 		line := scanner.Text()
-		if line[0] != '/' {
+		entry := parseMtabEntry(line)
+		if entry == nil || entry.Partition != part {
 			continue
 		}
-		fields := strings.Fields(line)
-		if fields[0] != part {
-			continue
-		}
-		opts := strings.Split(fields[3], ",")
 		ro := true
-		for _, opt := range opts {
+		for _, opt := range entry.Options {
 			if opt == "rw" {
 				ro = false
 				break
 			}
 		}
 
-		return fields[1], fields[2], ro
+		return entry.Mountpoint, entry.FilesystemType, ro
 	}
 	return "", "", true
+}
+
+type mtabEntry struct {
+	Partition      string
+	Mountpoint     string
+	FilesystemType string
+	Options        []string
+}
+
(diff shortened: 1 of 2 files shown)


block_linux.go
```diff
@@ -250,27 +250,62 @@ func PartitionInfo(part string) (string, string, bool) {
...
+	// tab and newline characters, encoded into the mtab entry line using their
+	// octal-to-string representations. From the GNU mtab man pages:
+	//
+	//   "Therefore these characters are encoded in the files and the getmntent
+	//   function takes care of the decoding while reading the entries back in.
+	//   '\040' is used to encode a space character, '\011' to encode a tab
+	//   character, '\012' to encode a newline character, and '\\' to encode a
+	//   backslash."
+	mp := fields[1]
+	mp = strings.Replace(mp, "\\011", "\t", -1)
```
thehowl: You can probably use a [`strings.Replacer`](https://golang.org/pkg/strings/#Replacer). Also, you'll need to add before other replaces a replace for <code>\`\\\\\`, \`\\`</code>

block_linux.go
```diff
@@ -250,27 +250,62 @@ func PartitionInfo(part string) (string, string, bool) {
...
+	Options        []string
+}
+
+func parseMtabEntry(line string) *mtabEntry {
+	// /etc/mtab entries for mounted partitions look like this:
+	// /dev/sda6 / ext4 rw,relatime,errors=remount-ro,data=ordered 0 0
+	if line[0] != '/' {
+		return nil
+	}
+	fields := strings.Fields(line)
```
thehowl: just to be sure that we don't panic later because of indexing outside of the slice, you should probably check `if len(fields) < 4`

block_linux.go
jaypipes: Ah, thanks for the hint about strings.Replacer :) I'd never heard of that...

block_linux.go
jaypipes: ack, good point. will add in a followup

block_linux.go
jaypipes: @thehowl thanks again for your review. fixed in https://github.com/jaypipes/ghw/commit/85a00ff9dc47d4e2e0ea159cfbd10495e682290a

PR #52, opened 2018-09-04
Added vet, fmtcheck, and fmt Makefile targets

retr0h:
Added a couple extra checks to run during `test`.  The fmtcheck
target looks the way it does, so returns an error if gofmt were
to reformat code.  Gofmt does not consider this an error.


jaypipes:
@retr0h if you feel like adding a similar patch to github.com/jaypipes/pcidb lib, please do! :)

PR #55, opened 2018-09-06
Move dep to single vendor dir

retr0h:
Move to a single vendor directory.  When moving to the dep tool,
I believe I incorrectly implemented it.  The ghwc package should
look at the parent vendor directory for dependencies just as all
other packages in this project.


jaypipes:
Yeah, I'm OK with this. Perhaps long term the goal should be to move ghwc into its own project. For now, having the cobra dependency on the primary library isn't too onerous.

Issue #56, opened 2018-09-18
Support alternate root mountpoints


chiptus:
Hi, 

First, thanks for this robust solution, I'm hoping to use it,
I have a specific use case where I want to run this library inside a docker container to get info about the host, and in that case, if understand correctly how your library works, the paths should be changed to where I bind the host folders to (e.g /host/proc). Do you think it's a doable change - to make the default paths changeable?

Thanks

chiptus:
it seems like the container share a lot of info with the host and thus `pci.listdevices` and on the container give almost the same info as the host.

jaypipes:
Hi @chiptus! Thanks for trying out ghw :) I'm a little uncertain what you mean by changing the default folders. Maybe you could elaborate on your request?

The Travis CI system which tests each commit to ghw actually runs in a container, so I'm pretty sure that whether or not the ghw is installed in a container, a virtual machine, or on baremetal should not matter. Certainly, if it does matter, then I want to fix things up!

Can you give me some details regarding what *type* of container you are using? Is it a privileged (system) container with a full root filesystem? Is it a simple application container with only a couple mountpoints (for /etc and container-local applications mountpoints)?

Is there a specific error you are seeing when you run ghw? If so, please paste it here and I'll see what I can do about making ghw more tolerant to the kinds of application containers you're using.

Thanks!
-jay

deviantony:
Hi @jaypipes 

I'm working with @chiptus on this. Our goal would be to use the `ghw` library in a binary running inside a container to retrieve information about the host where the container is running.

Our assumption was that `ghw` reads its information from specific files on the filesystem (such as `/proc/cpuinfo` for CPU information). 

Thus, our idea was to bind-mount the `/` filesystem of the host into `/host` in the container (`docker run -v /:/host:ro mycontainer ...`) and to be able to customize the "base path" where `ghw` would retrieve its information (`/host` in this case, to be able to query `/host/proc/cpuinfo` for example).

We're not using any privileged container, it's basically a golang binary running inside a `scratch` image, root access should not be a problem.

jaypipes:
Ah, @deviantony gotcha. Alrighty, lemme push something up later today that you can pull and test with, eh?

deviantony:
@jaypipes sounds good ! Will be waiting for the update :-)

jaypipes:
[...]

PR #58, opened 2018-09-18
Allow overriding root path/bindmount

jaypipes:
If a host filesystem has been bind-mounted to a particular directory
within an application container (say, '/host'), we want to inform `ghw`
to use that bind-mount as its root directory when constructing paths to
certain files like /proc/cpuinfo or /etc/mtab.

This patch adds the ability for the user to set the GHW_CHROOT environ
variable to the value of this alternate root mountpoint, allowing `ghw`
to find important files from within an application container that has
bind-mounted a host filesystem.

Issue #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ information about the host computer:
 * [PCI](#pci)
 * [GPU](#gpu)
 
+**NOTE**: The default root mountpoint that `ghw` uses when looking for
+information about the host system is `/`. So, for example, when looking up CPU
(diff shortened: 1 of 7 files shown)


chiptus:
This is still a constant

https://github.com/jaypipes/ghw/blob/master/pci_linux.go#L19

jaypipes:
ack, good catch @chiptus, I'll remove shortly.

chiptus:
Great, I will test tomorrow

Issue #57, opened 2018-09-18
block is not consistent with `df`


chiptus:
For some reason, while I have only 256GB, running the block example code shows I have 2TB.

running `df -h`:
![screenshot from 2018-09-18 13-10-48](https://user-images.githubusercontent.com/1381655/45680728-4e433e00-bb44-11e8-8e7c-584f3bcd072e.png)

running the following:
```
ackage main

import (
	"fmt"

	"github.com/jaypipes/ghw"
)

func main() {
	block, err := ghw.Block()
	if err != nil {
		fmt.Printf("Error getting block storage info: %v", err)
	}

	fmt.Printf("%v\n", block)

	for _, disk := range block.Disks {
		fmt.Printf(" %v\n", disk)
		for _, part := range disk.Partitions {
			fmt.Printf("  %v\n", part)
		}
	}
}
```

![screenshot from 2018-09-18 13-12-20](https://user-images.githubusercontent.com/1381655/45680778-8185cd00-bb44-11e8-8a3f-90caff40d9b4.png)

jaypipes:
Hi @chiptus, that's odd. It seems to be indicating that ghw thinks the output of your first disk's `String()` method is "204848114432". I have no idea why that would be, considering `ghw.Disk.String()`'s implementation looks like this:

```go
func (d *Disk) String() string {
	vendor := ""
	if d.Vendor != "" {
		vendor = "  " + d.Vendor
	}
	serial := ""
	if d.SerialNumber != "" {
		serial = " - SN #" + d.SerialNumber
	}
	sizeStr := "unknown"
	if d.SizeBytes > 0 {
		size := d.SizeBytes
		unit, unitStr := unitWithString(int64(size))
		size = uint64(math.Ceil(float64(size) / float64(unit)))
		sizeStr = fmt.Sprintf("%d%s", size, unitStr)
	}
	return fmt.Sprintf(
		"/dev/%s (%s) [%s]%s%s",
		d.Name,
		sizeStr,
		d.BusType,
		vendor,
		serial,
	)
}
```

In order to diagnose what's going on, could you paste the output of the following on your machine please?

```
cat /sys/block/sda/queue/physical_block_size
cat /sys/block/sda/size
cat /sys/block/sda/vendor
```

Thanks in advance!
-jay

chiptus:
- `cat /sys/block/sda/queue/physical_block_size`: 4096
- `cat /sys/block/sda/size`: 500118192
[...]

PR #60, opened 2018-09-20
Correct improper Linux block device size calc

jaypipes:
Linux hard-codes the block device sector size to 512, and when
determining the total size of a disk, we multiply this 512 hard-coded
number by the value in /sys/block/$device/size.

We were erroneously calculating the block device's size by multiplying
the physical block size returned from
/sys/block/$device/queue/physical_block_size with the value in
/sys/block/$device/size.

Issue #57


jaypipes:
@chiptus please let me know if this pull request fixes the issue you reported in Issue #57. Thanks!

chiptus:
great, it works.

Thanks!

PR #64, opened 2018-09-28
Collect more disk information

jpo-joyent:
This PR adds code to also collect:
- which NUMA node a disk is local to (useful when trying to balance IO)
- the disk's bus path (useful when trying to know where to unplug the thing from)
- the disk's model (useful in conjunction with the serial number)

Also fixes a bug where serial numbers would be listed as #unknown instead of being omitted

One downside is we re-read and re-parse the udev db (a small flat file) for each property lookup. I could cache it, but it would make the code uglier. Let me know if you'd like me to.

If you need me to break these out into separate PRs, just let me know.

### DIFF
--- a/block.go
+++ b/block.go
@@ -16,6 +16,7 @@ type Disk struct {
 	SizeBytes              uint64
 	PhysicalBlockSizeBytes uint64
 	BusType                string
+	BusPath                string
 	Vendor                 string
 	SerialNumber           string
 	Partitions             []*Partition
@@ -79,10 +80,11 @@ func (d *Disk) String() string {
 		sizeStr = fmt.Sprintf("%d%s", size, unitStr)
 	}
 	return fmt.Sprintf(
-		"/dev/%s (%s) [%s]%s%s",
+		"/dev/%s (%s) [%s @ %s]%s%s",
 		d.Name,
 		sizeStr,
 		d.BusType,
+		d.BusPath,
 		vendor,
 		serial,
 	)
--- a/block.go
+++ b/block.go
@@ -20,6 +20,7 @@ type Disk struct {
 	Vendor                 string
 	SerialNumber           string
 	Partitions             []*Partition
+	NumaNodeId             int
 }
 
 type Partition struct {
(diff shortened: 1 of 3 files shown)


jpo-joyent:
This PR is on top of https://github.com/jaypipes/ghw/pull/63

jaypipes:
@jpo-joyent Hi! I've already merged #63 so would you mind rebasing to master and re-pushing to this PR? That would remove the udevInfo() code that is contained in here.

block_linux.go
```diff
@@ -74,53 +89,83 @@ func DiskVendor(disk string) string {
...
+	udevInfo := make(map[string]string)
+	for _, udevLine := range strings.Split(string(udevBytes), "\n") {
+		if strings.HasPrefix(udevLine, "E:") {
+			if s := strings.SplitN(udevLine[2:], "=", 2); len(s) == 2 {
+				udevInfo[s[0]] = s[1]
+			}
+		}
+	}
+	return udevInfo, nil
+}
```
jaypipes: The above code is already in master:

https://github.com/jaypipes/ghw/blob/dc696826c46f9a64b1e4c3b7304b0c68e8042a8f/block_linux.go#L77-L100

Could you rebase to master to remove this from this PR?

block_linux.go
```diff
@@ -74,53 +89,83 @@ func DiskVendor(disk string) string {
...
-			return lname[pos+1:]
-		}
+
+	// There are two serial number keys, ID_SERIAL and ID_SERIAL_SHORT
+	// The non-_SHORT version often duplicates vendor information collected elsewhere, so use _SHORT.
+	if serial, ok := info["ID_SERIAL_SHORT"]; ok {
+		return serial
+	}
+	return UNKNOWN
+}
```
jaypipes: The above code is also already in master...

block.go
```diff
@@ -16,8 +16,12 @@ type Disk struct {
 	SizeBytes              uint64
 	PhysicalBlockSizeBytes uint64
 	BusType                string
+	BusPath                string
+	NumaNodeId             int
```
jaypipes: Please name the attribute "NUMANodeID". Per Golang naming convention best practices, acronyms should remain capitalized:

https://talks.golang.org/2014/names.slide#5

I need to go through `ghw` and double-check that all the field names follow these conventions...

block.go
```diff
@@ -16,8 +16,12 @@ type Disk struct {
...
 	Vendor                 string
+	Model                  string
 	SerialNumber           string
+	WorldWideName          string
```
jaypipes: I could support naming this field just "WWN" since that seems to be what most folks would expect/look for?

block.go
jpo-joyent: Noted. Thanks!

block.go
jpo-joyent: Sounds good to me.

block_linux.go
jpo-joyent: GitHub's UI is just stale here. This is being pulled in from as part of commit 55f8af4 which is already merged, so a further merge would only have one copy. Rebased on top of your merge commit anyway though.

block_linux.go
jpo-joyent: Right, same parent commit. Merge would be NOOP.

jpo-joyent:
Suggestions applied.

jaypipes:
Awesome, thanks so much @jpo-joyent! :)

PR #67, opened 2018-10-11
Build on non-linux platforms

jpo-joyent:
Not collecting useful information (yet), but at least doesn't break builds for multi-platform projects which include this as a dependency.

### DIFF
--- a/block_stub.go
+++ b/block_stub.go
@@ -13,7 +13,7 @@ import (
 )
 
 func blockFillInfo(info *BlockInfo) error {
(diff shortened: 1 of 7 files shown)


jaypipes:
Rock on, @jpo-joyent :)

I'll work on adding TravisCI jobs for !linux platforms and see about starting to exercise this code in the gates.

Thanks very much for your contributions! (I wish I knew your name so I could refer to you as more than just "jpo-joyent"!)

PR #69, opened 2018-10-11
Stop exporting linux-specific paths

jpo-joyent:
These make no sense on non-linux platforms and wouldn't exist in `*_otherOS.go` files, so remove them now before anybody starts accidentally using them to avoid breaking backwards compat when adding support for other OSes.

### DIFF
--- a/path_linux.go
+++ b/path_linux.go
@@ -53,6 +53,14 @@ func pathSysBusPciDevices() string {
 	return filepath.Join(pathRoot(), "sys", "bus", "pci", "devices")
 }
 
+func pathSysClassDrm() string {
+	return filepath.Join(pathRoot(), "sys", "class", "drm")
+}
+
+func pathSysClassNet() string {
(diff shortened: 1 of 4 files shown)


jpo-joyent:
Force-pushed updated commit including `RegexNVMe*`

path_linux.go
```diff
@@ -53,6 +53,14 @@ func pathSysBusPciDevices() string {
 	return filepath.Join(pathRoot(), "sys", "bus", "pci", "devices")
 }
 
+func pathSysClassDrm() string {
+	return filepath.Join(pathRoot(), "sys", "class", "drm")
+}
+
+func pathSysClassNet() string {
+	return filepath.Join(pathRoot(), "sys", "class", "net")
+}
```
jaypipes: Cool, thanks for moving these. I'd obviously missed these when I did the new path_linux.go module. ++

Issue #86, opened 2018-11-11
Allow configuration options to be passed to module constructors


chiptus:
It seems a bit weird to set GHW_CHROOT as env var. we might have a few containers that each of them has a different CHROOT, so would like to have the option to set it for each invocation. an env var is seems too global. Maybe you can have a options object that is passed to each of the root methods (like `ghw.CPU()`, and then check if it's missing the chroot then check for the env var and if it's missing set it to the default root.

What do you say?

retr0h:
> It seems a bit weird to set GHW_CHROOT as env var. we might have a few containers that each of them has a different CHROOT

Seems like you are still having to change something regardless.  Either the env var or the proposed cli option.

jaypipes:
Hi @chiptus! Hope you had a good weekend! :)

I need to better understand your use case, I think, but as far as I can tell, if you had multiple containers that you wanted to set the `GHW_CHROOT` to a different value, you'd start those containers with different `GHW_CHROOT` environ variables.

For instance, if I was using Docker, I'd use the `-e GHW_CHROOT=<some path>` flag when executing `docker run`.

Am I misunderstanding things?

deviantony:
Hey @jaypipes @retr0h 

In my opinion it does not really make sense to use environment variables to alter the behavior of a library. I think that it would be cleaner to just pass a parameter to the "constructor/accessor" of the library instead. 

Why would we want to alter the entire execution context (via an environment variable) just to alter the execution of a specific part of a program? Using an environment variable to set an option/configuration makes more sense in the context of using a program, not a library IMO.

jaypipes:
@deviantony I guess I disagree with you on this one -- or at least, the UNIX way is to allow software libraries (as well as programs) to be configured via environ variables. Plenty of C libraries use environs variables like `LANG` or `PATH` or `LC_TIME` to alter their output or behaviour.

Having said that, I'm certainly not opposed to doing what you suggest: having an `Options` struct be passable to the various module constructors to make configuration even more flexible.

If I'm reading you correctly, I think you're suggesting something akin to this?

```go
type Options struct {
    Chroot string
}

func CPUWithOptions(opts *Options) (*CPUInfo, error) {
...
}

// envOptions returns a point to an Options struct that has been
// filled in with environs variables (such as GHW_CHROOT)
func envOptions() (*Options) {
    chrootEnv := os.Getenv("GHW_CHROOT")
    if chrootEnv == "" {
        chrootEnv = "/"
    }
[...]

PR #88, opened 2018-11-14
Support specifying options in module constructors

jaypipes:
Each module (CPU(), Memory(), Network(), etc...) now supports calling
the main module constructor with zero or more pointers to WithOption
structs. There is only a single option (the CHROOT value for path
construction) but I've designed it so the system is flexible for future
additional options without needing to change the calling interface.

To specify a chroot override for a particular module, use the
WithChroot() function, like so:

```go
mem, err := ghw.Memory(WithChroot("/host"));
```

Issue #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,15 +34,28 @@ information about the host computer:
 * [PCI](#pci)
 * [GPU](#gpu)
 
-**NOTE**: The default root mountpoint that `ghw` uses when looking for
-information about the host system is `/`. So, for example, when looking up CPU
-information on Linux, `ghw` will attempt to parse the `/proc/cpuinfo` file. If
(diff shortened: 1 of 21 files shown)


jaypipes:
@deviantony @chiptus if you have the opportunity to pull this commit (78e8fd3) and test it in your environment before I merge, that would be great. :)

deviantony:
@jaypipes we will test this PR in our env, thanks !

chiptus:
@jaypipes, thanks for all your work, we really appreciate it.

our problem is mostly with pcidb, so we can't test it until the new version of pcidb is in.

jaypipes:
Ah, OK. Let me do the remaining work for pcidb, cut a release and then use the new release in ghw.

PR #95, opened 2018-12-03
Adjust *_stub.go to new internal context API

jpo-joyent:
This fixes the build on !linux again (tested on macOS), although still doesn't report anything useful there.


jpo-joyent:
Previously: https://github.com/jaypipes/ghw/pull/67

Issue #39, opened 2018-08-08
Unable to use ghw on MacOS


jaypipes:
Everything written for ghw so far is Linux-centric (since that's what I work with).

It would be great to have ghw work on MacOS/Darwin. I think I can do that with minimal changes, since OSX is basically a Linux/BSD-like system

ykcab:
referencing to the last response #8, I will do a pull request once I am certain what I have works for macOS.

jpo-joyent:
Disks done (w/ plist dep) in https://github.com/jaypipes/ghw/pull/97 / https://github.com/jaypipes/ghw/pull/98.

jpo-joyent:
Not sure how stable it actually is. Let's hope you're right :)

There's never been a single release, so I pinned the arbitrary latest commit in Gopkg.toml

jaypipes:
Closing this out. We can have individual issues for support on MacOSX for each specific submodules in ghw (network, cpu, etc...)

PR #97, opened 2018-12-11
Implement block device support for macOS

jpo-joyent:
This partially addresses https://github.com/jaypipes/ghw/issues/39

### DIFF
--- a/block_darwin.go
+++ b/block_darwin.go
@@ -94,35 +94,35 @@ type ioregPlist struct {
 	VendorName   string `plist:"Vendor Name"`
 }
 
-func (ctx *context) getDiskUtilListPlist() (diskUtilListPlist, error) {
+func (ctx *context) getDiskUtilListPlist() (*diskUtilListPlist, error) {
 	out, err := exec.Command("diskutil", "list", "-plist").Output()
 	if err != nil {
-		return diskUtilListPlist{}, errors.Wrap(err, "diskutil list failed")
+		return nil, errors.Wrap(err, "diskutil list failed")
 	}
 
 	var data diskUtilListPlist
 	if _, err := plist.Unmarshal(out, &data); err != nil {
-		return diskUtilListPlist{}, errors.Wrap(err, "diskutil list plist unmarshal failed")
+		return nil, errors.Wrap(err, "diskutil list plist unmarshal failed")
 	}
 
-	return data, nil
+	return &data, nil
 }
 
-func (ctx *context) getDiskUtilInfoPlist(device string) (diskUtilInfoPlist, error) {
+func (ctx *context) getDiskUtilInfoPlist(device string) (*diskUtilInfoPlist, error) {
 	out, err := exec.Command("diskutil", "info", "-plist", device).Output()
 	if err != nil {
-		return diskUtilInfoPlist{}, errors.Wrapf(err, "diskutil info for %q failed", device)
+		return nil, errors.Wrapf(err, "diskutil info for %q failed", device)
 	}
 
 	var data diskUtilInfoPlist
 	if _, err := plist.Unmarshal(out, &data); err != nil {
-		return diskUtilInfoPlist{}, errors.Wrapf(err, "diskutil info plist unmarshal for %q failed", device)
+		return nil, errors.Wrapf(err, "diskutil info plist unmarshal for %q failed", device)
 	}
 
-	return data, nil
+	return &data, nil
 }
 
-func (ctx *context) getIoregPlist(ioDeviceTreePath string) (ioregPlist, error) {
+func (ctx *context) getIoregPlist(ioDeviceTreePath string) (*ioregPlist, error) {
 	elems := strings.Split(ioDeviceTreePath, "/")
 	name := elems[len(elems)-1]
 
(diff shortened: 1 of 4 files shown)


jaypipes:
@jpo-joyent OK, sorry for delay on this. So, I'm cool with the proposed library (plist) inclusion. Looks like a solid lib, so I'm good with it, especially since you've pinned to a specific commit (thank you!).

However, before merging this code, I want to merge a couple changes that fix up the tests for OSX and Windows so that we have some test stubs to fill out. Give me a couple more days to get that done.

Thanks!
-jay

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+
+package ghw
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
```
jaypipes: You will need to add this dependency to Gopkg.toml as well. Alternately, consider just using `fmt.Errorf()` everywhere instead of the github.com/pkg/errors library.

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+type ioregPlist struct {
+	// there's a lot more than just this...
+	ModelNumber  string `plist:"Model Number"`
+	SerialNumber string `plist:"Serial Number"`
+	VendorName   string `plist:"Vendor Name"`
+}
+
+func (ctx *context) getDiskUtilListPlist() (diskUtilListPlist, error) {
```
jaypipes: For all of these methods, please change the return value to (*diskUtilListPlist, error). Instead of returning a "blank" `diskUtilListPlist{}` like on lines 100 and 105, instead return `nil, errors.Wrap(err, ...)`

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+	if _, err := plist.Unmarshal(out, &data); err != nil {
+		return diskUtilInfoPlist{}, errors.Wrapf(err, "diskutil info plist unmarshal for %q failed", device)
+	}
+
+	return data, nil
+}
+
+func (ctx *context) getIoregPlist(ioDeviceTreePath string) (ioregPlist, error) {
+	elems := strings.Split(ioDeviceTreePath, "/")
+	name := elems[len(elems)-1]
```
jaypipes: For the above, please use:

```
name := filepath.Base(ioDeviceTreePath)
```

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+	if len(data) != 1 {
+		err := errors.Errorf("ioreg unmarshal resulted in %d I/O device tree nodes (expected 1)", len(data))
+		return ioregPlist{}, err
+	}
+
+	return data[0], nil
+}
+
+func (ctx *context) makePartition(disk, s diskOrPartitionPlistNode, isAPFS bool) (*Partition, error) {
+	if s.Size < 0 {
```
jaypipes: If you made diskOrPartitionPlistNode.Size a `uint64` this wouldn't be necessary.

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+		fmt.Fprintln(os.Stderr, err.Error())
+		return err
+	}
+
+	info.TotalPhysicalBytes = 0
+	info.Disks = make([]*Disk, 0, len(listPlist.AllDisksAndPartitions))
+	info.Partitions = []*Partition{}
+
+	for _, disk := range listPlist.AllDisksAndPartitions {
+		if disk.Size < 0 {
```
jaypipes: Ditto. If `Size` was a `uint64` this wouldn't be necessary.

block_darwin.go
```diff
@@ -0,0 +1,250 @@
...
+			return errors.Errorf("invalid size %q of disk %q", disk.Size, disk.DeviceIdentifier)
+		}
+
+		infoPlist, err := ctx.getDiskUtilInfoPlist(disk.DeviceIdentifier)
+		if err != nil {
+			return err
+		}
+		if infoPlist.DeviceBlockSize < 0 {
```
jaypipes: same with `DeviceBlockSize`.

block_darwin.go
jpo-joyent: Done.

block_darwin.go
jpo-joyent: Fixed, except using path.Base instead of filepath.Base, since the iokit device tree is its own thing and has nothing to do with the filesystem.

block_darwin.go
jpo-joyent: Except I don't control the input, it's not open source (as far as I can tell) so I can't determine if it's possible to ever receive a negative value (I've never seen one, and can't think of why there would be, but that doesn't mean there never would be). This way makes catching the error explicit and the root cause obvious rather than a "cannot unmarshal to type" or such.

block_darwin.go
jpo-joyent: See above.

block_darwin.go
jpo-joyent: See above.

block_darwin.go
jpo-joyent: I don't feel strongly about this though. If you still want me to change it, I will.

jaypipes: @jpo-joyent yeah, you make a good point about the relative obscurity of the error that would be produced if the (closed-source) plist response ever contained a negative size. I suppose that's the suck about dealing with closed-source software :)

OK, I'm cool leaving this as-is.

PR #98, opened 2018-12-11
ghwc: Gather only the info required in each subcommand

jpo-joyent:
This makes it possible to actually produce useful output from ghwc on platforms which only implement a subset of ghw's functionality. Right now this is the case for macOS, which so far only supports querying block devices.


jaypipes:
Yeah, `ghwc` started out life as merely an example of how to use the `ghw` library, so not much thought was put into user experience. :) Thanks for cleaning this up, @jpo-joyent!

PR #99, opened 2018-12-11
ghwc: Lexically sort subcommand invocation in root cmd

jpo-joyent:
This matches the order in which cobra lists the subcommands too.


jaypipes:
Thanks, Jean-Philippe!

PR #100, opened 2018-12-14
net_linux : Add detection to know if nic is on PCI BUS

mikefaille:
I need it to find physical interface on my server; especially ones on PCI BUS
If nic is on pci bus, print nic according this info
Update documentation according nic pci detection

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,7 @@ Each `ghw.NIC` struct contains the following fields:
 * `ghw.NIC.MacAddress` is the MAC address for the NIC, if any
 * `ghw.NIC.IsVirtual` is a boolean indicating if the NIC is a virtualized
   device
+* `ghw.NIC.IsOnPciBus` is a boolean indicating if the NIC is on PCI bus
 * `ghw.NIC.Capabilities` is an array of pointers to `ghw.NICCapability` structs
   that can describe the things the NIC supports. These capabilities match the
   returned values from the `ethtool -k <DEVICE>` call on Linux
@@ -485,14 +486,14 @@ net (3 NICs)
    - tx-sctp-segmentation
    - tx-esp-segmentation
    - tx-vlan-stag-hw-insert
- enp58s0f1
+ enp58s0f1 (pci)
   enabled capabilities:
    - rx-checksumming
    - generic-receive-offload
    - rx-vlan-offload
    - tx-vlan-offload
    - highdma
- wlp59s0
+ wlp59s0 (pci)
   enabled capabilities:
    - scatter-gather
    - tx-scatter-gather
--- a/net.go
+++ b/net.go
@@ -20,6 +20,7 @@ type NIC struct {
 	Name         string
 	MacAddress   string
 	IsVirtual    bool
+	IsOnPciBus   bool
 	Capabilities []*NICCapability
 }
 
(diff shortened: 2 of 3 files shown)


pwFoo:
Dectect NIC to initialize network after booting a minimal system would be great!
Would need it not only for PCI devices.

mikefaille:
@pwFoo Maybe, we can detect virtual devices more easily using something like :  

```
ls /sys/class/net -l | grep devices/virtual/net/
lrwxrwxrwx 1 root root 0 déc 15 11:02 br-0a5a2e80b26b -> ../../devices/virtual/net/br-0a5a2e80b26b
lrwxrwxrwx 1 root root 0 déc 15 11:02 br-6847a3df4828 -> ../../devices/virtual/net/br-6847a3df4828
lrwxrwxrwx 1 root root 0 déc 15 11:02 br-a9a72a125b03 -> ../../devices/virtual/net/br-a9a72a125b03
lrwxrwxrwx 1 root root 0 déc 15 11:02 docker0 -> ../../devices/virtual/net/docker0
lrwxrwxrwx 1 root root 0 déc 15 11:02 docker_gwbridge -> ../../devices/virtual/net/docker_gwbridge
lrwxrwxrwx 1 root root 0 déc 15 11:02 lo -> ../../devices/virtual/net/lo
lrwxrwxrwx 1 root root 0 déc 15 11:02 veth2be86f9 -> ../../devices/virtual/net/veth2be86f9
lrwxrwxrwx 1 root root 0 déc 15 11:02 virbr0 -> ../../devices/virtual/net/virbr0
lrwxrwxrwx 1 root root 0 déc 15 11:02 virbr0-nic -> ../../devices/virtual/net/virbr0-nic
```

By excluding virtual interfaces (ones from previous list), we can always find physical nic (I think ). 

The trouble here is there is a conflict about `virtual` definition on actual code. 
https://github.com/jaypipes/ghw/blob/master/net_linux.go#L62
Here, we define `virtio` nic as `virtual` ones.

My suggestion then is creating a new major release that break previous behaviour : I think IsVirtual() must be true for NIC giving `true` when their symlink from `ls -l /sys/class/net/*` contain ` devices/virtual/net/`

jaypipes:
> My suggestion then is creating a new major release that break previous behaviour : I think IsVirtual() must be true for NIC giving `true` when their symlink from `ls -l /sys/class/net/*` contain ` devices/virtual/net/`

I would support changing the code that determines if a NIC is virtual. I do not support, however, the code that is in this patch at the moment, that looks for "pci" in the device path and adds another boolean field to the `NIC` struct.

@mikefaille I'm a little unsure what your end goal is here. Perhaps you could elaborate? Are you just looking for the low-level hardware (or software, in the case of virtual NICs) bus that handles the NIC? Or are you running into a problem where (as @pwFoo mentions), a virtual NIC is mistakenly being reported as a non-virtual NIC because the code that determines "virtual-ness" is looking for the wrong thing in the device path?

jaypipes:
> I can implement that if it's ok. It will works for me.

Are you asking if you can change the implementation of the IsVirtual() function? Yes, that would be fine.

mikefaille:
> > I can implement that if it's ok. It will works for me.
> 
> Are you asking if you can change the implementation of the IsVirtual() function? Yes, that would be fine.

Sharp, I'll do it !

mikefaille:
I close this actual pull request in favour of this one : https://github.com/jaypipes/ghw/pull/100

PR #101, opened 2018-12-18
correct detection of virtual interfaces on linux

mikefaille:
To detect interfaces like :
- veth
- tun
- bridge
- and, Virtio
- etc;

I suggest to capture sys network interface class path contain
'devices/virtual/net' on symlink target.


mikefaille:
As mentioned on previous pull request reference, it would help to detect physical hardware too.

PR #104, opened 2019-01-13
merge darwin network support

marcusguttenplan:


marcusguttenplan:
Merged

jaypipes:
Hi @marcusguttenplan :) This looks like it was kind of a work in progress... is that right? Were you looking for some assistance on things?

Best,
-jay

marcusguttenplan:
HUGE work in progress, apologies for any notification spam Jay. Still working through reversing your great lib, hoping to port a `net`-based implementation of some similar network tools that I've been using on darwin.

jaypipes:
Absolutely no need to apologize, Marcus! :) Let us know if you'd like any assistance in what you're doing. I know at least @jpo-joyent and a couple others are working on Darwin-related support so you might want to team up with them!

jpo-joyent:
Implementing further darwin support is not a priority for me at the moment, but I'd be glad to review & test any proposed patches for it once you consider them ready.

My only suggestion so far from reading this patch is that you try to see if you can gather the desired information out of `ioreg` (which is necessarily available by default on all macos systems) rather than relying on the 3rd-party `ethtool` which is unlikely to be present on most machines.

PR #105, opened 2019-01-27
Add Virtio as BusType

blp1526:
I add the Virtio BusType for KVM users.

https://www.linux-kvm.org/page/Virtio
> Disk will show up as /dev/vd[a-z][1-9]


jaypipes:
Hi! Thanks very much for the contribution! :)

PR #107, opened 2019-01-29
Use GitHub url for howett.net/plist

jpo-joyent:
Because howett.net keeps going down :(

See https://github.com/DHowett/go-plist/issues/38


jaypipes:
:+1: thanks!

PR #115, opened 2019-01-31
dep: s/constraint/override/ for transitive dep

jpo-joyent:
otherwise:
```
$ dep ensure
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  gopkg.in/yaml.v2

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,62 +2,86 @@
 
 
 [[projects]]
+  digest = "1:b13707423743d41665fd23f0c36b2f37bb49c30e94adb813319c44188a51ba22"
   name = "github.com/ghodss/yaml"
   packages = ["."]
+  pruneopts = ""
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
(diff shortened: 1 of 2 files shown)


jaypipes:
Hey Jean Paul, thanks for submitting this PR.

I believe I messed things up by originally doing a `dep ensure -add gopkg.in/yaml.v2@v2.2.0` but then when I ran into issues with not being able to define a custom MarshalYAML function for a couple types, I went with the ghodss/yaml library and did a `dep ensure -add github.com/ghodss/yaml@v1.0.0`.

Thanks for cleaning up my mistakes!
-jay

PR #116, opened 2019-01-31
Map macOS bus protocols to BusType enum

jpo-joyent:
In macOS, the bus type is a string that comes from the IOKit BusProtocol field, whose legal values are not documented, and whose source is not available. The only value I've seen is `PCI-Express`, but this does not mean other things are not possible, just that I have not seen them so far.

BusType now being a string let me just pass them through and let the caller have something useful in cases I have not anticipated, but now I don't have a clean option. Always passing `BUS_TYPE_UNKNOWN` seems less than ideal...

### DIFF
--- a/block_darwin.go
+++ b/block_darwin.go
@@ -178,6 +178,15 @@ func (ctx *context) makePartition(disk, s diskOrPartitionPlistNode, isAPFS bool)
 	}, nil
 }
 
+func busTypeFromBusProtocol(busProtocol string) BusType {
+	switch busProtocol {
+	case "PCI-Express":
+		return BUS_TYPE_NVME // is this even correct?
+	default:
+		return BUS_TYPE_UNKNOWN
+	}
+}
+
 func (ctx *context) blockFillInfo(info *BlockInfo) error {
 	listPlist, err := ctx.getDiskUtilListPlist()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


block_darwin.go
```diff
@@ -178,6 +178,15 @@ func (ctx *context) makePartition(disk, s diskOrPartitionPlistNode, isAPFS bool)
 	}, nil
 }
 
+func busTypeFromBusProtocol(busProtocol string) BusType {
+	switch busProtocol {
+	case "PCI-Express":
+		return BUS_TYPE_NVME // is this even correct?
```
jaypipes: hmm, from what I can tell, modern Macs use SATA drives. They used to use SCSI drives. Perhaps we need to rethink how we describe the "bus type" for block devices and instead indicate the storage interface (SATA vs SCSI, etc)... thoughts?

jpo-joyent: I'm not sure, but I think my machine actually does use some (proprietary?) NVMe storage device. I'm only inferring this from the fact that it shows up under the "NVMExpress" section of "System Information", and [this article from a random search](https://beetstech.com/blog/apple-proprietary-ssd-ultimate-guide-to-specs-and-upgrades) seems to confirm.

Beyond that, I don't know, and this random new-ish MacBook Pro is the only Apple hardware I have to test on. My implementation of BusType is really just a best-effort to populate the field with _something_ for parity with the reports from Linux.

jaypipes:
Please check out #119 for the approach I came up with for differentiating drive type and storage controller. You'll note that BusType in that PR for Macs is hard-coded to BUS_TYPE_PCI and the StorageController is set to STORAGE_CONTROLLER_NVME when the IONVMeController string is found in the DeviceTreePath in the plist output.

jaypipes:
I believe this PR can be closed now that #119 is merged.

PR #119, opened 2019-02-10
differentiate drive type and storage controller

jaypipes:
This patch attempts to bring some clarity to the block device module's
`ghw.Disk` struct with regards to the storage controller and drive type.
We add a new enumeration of drive types to `ghw`:

* DRIVE_TYPE_HDD - hard disk drive
* DRIVE_TYPE_FDD - floppy disk drive
* DRIVE_TYPE_ODD - optical disk drive
* DRIVE_TYPE_SSD - solid-state drive

Along with a new `ghw.StorageController` enumeration which describes the
controller/driver used for translating block reads/writes to the
underlying hardware, which the following enum values:

* STORAGE_CONTROLLER_IDE - integrated device electronics
* STORAGE_CONTROLLER_SCSI - small computer system interface
* STORAGE_CONTROLLER_NVME - non-volatile memory express
* STORAGE_CONTROLLER_VIRTIO - libvirt/virtio storage driver

The `ghw.Disk.BusType` field is deprecated and no longer serialized into
JSON/YAML output.

Issue #117

PR #120, opened 2019-02-11
Fix build on macOS again

jpo-joyent:
Some macOS CI would be awesome ;)


jpo-joyent:
Note, this is proposing to merge into `issue117` (the HEAD for https://github.com/jaypipes/ghw/pull/119) not `master`.

PR #125, opened 2019-02-17
Added USB-STORAGE field for /dev/sd* devices

chatenilesh:


block.go
```diff
@@ -16,20 +16,22 @@ import (
...
-	DRIVE_TYPE_HDD               // Hard disk drive
-	DRIVE_TYPE_FDD               // Floppy disk drive
-	DRIVE_TYPE_ODD               // Optical disk drive
-	DRIVE_TYPE_SSD               // Solid-state drive
+	DRIVE_TYPE_UNKNOWN     DriveType = iota
+	DRIVE_TYPE_HDD                   // Hard disk drive
+	DRIVE_TYPE_FDD                   // Floppy disk drive
+	DRIVE_TYPE_ODD                   // Optical disk drive
+	DRIVE_TYPE_SSD                   // Solid-state drive
+	DRIVE_TYPE_USB_STORAGE           // USB flash drive, SSD/HDD by USB
```
jaypipes: I'm uncomfortable adding a "USB_STORAGE" drive type. I don't believe that USB is actually a drive type; rather, USB is a bus and set of storage/transfer protocol standards. The drive type is either HDD, SSD or ODD (since there are some CD/DVD optical drives that are connected via USB).

block_linux.go
```diff
@@ -413,6 +417,17 @@ func (ctx *context) diskIsRotational(devName string) bool {
 	return contents == 1
 }
 
+func (ctx *context) diskIsUSBConnected(devName string) bool {
+	if info, err := ctx.udevInfo(devName); err == nil {
+		if value, ok := info["ID_BUS"]; ok {
+			//shall we use this?
+			//ID_USB_DRIVER=usb-storage
```
jaypipes: So, this actually got me thinking... what about adding a new field to the `ghw.Disk` struct called `StorageDriver` and put the software (or firmware) storage driver name in that field? This storage driver information might be useful for callers of `ghw.Block()` and means we would prevent mixing the storage driver concept with the (separate) concept of a drive type.

Nilesh, what are your thoughts on that? Do you think you could take on writing a diskStorageDriver() function that queried udev (as above) and/or sysfs on Linux to determine the storage driver for a disk?

jaypipes: @jpo-joyent what do you think about adding a `ghw.Disk.StorageDriver` field?

chatenilesh: Sure :)
I will look into it and make the changes

jpo-joyent: > @jpo-joyent what do you think about adding a `ghw.Disk.StorageDriver` field?

It's unclear to me what this means.

> and put the software (or firmware) storage driver name in that field?

Device firmware is useful to report, but software? Like whether it attaches via `ahci(4)` or whatever? That's rather non-portable and has more to do with your host OS than your hardware - and to make matters worse, device drivers often layer upon each other in a tree manner reflecting the bus topology.

Unless I misunderstand the intention, such a field seems confusing and redundant IMHO.

jaypipes: > > @jpo-joyent what do you think about adding a `ghw.Disk.StorageDriver` field?
> 
> It's unclear to me what this means.

I was referring to @chatenilesh's comment in his submitted code:

```go
//shall we use this?
//ID_USB_DRIVER=usb-storage
```

and asking whether it would be useful to have "driver" information included in the `ghw.Disk` struct's fields.

> > and put the software (or firmware) storage driver name in that field?
> 
> Device firmware is useful to report, but software? Like whether it attaches via `ahci(4)` or whatever? That's rather non-portable and has more to do with your host OS than your hardware - and to make matters worse, device drivers often layer upon each other in a tree manner reflecting the bus topology.
> 
> Unless I misunderstand the intention, such a field seems confusing and redundant IMHO.

OK, that's the opinion I was looking for, thank you @jpo-joyent :)

chatenilesh: I am using BusPath and and checking if the device as USB-STORAGE.
Will close the PR as these changes don't seem to be useful.

PR #130, opened 2019-05-29
Added support for parsing AWS xvd prefixed block devices

mvisonneau:
The current parser ignores `/dev/xvd*` block devices :( this should fix it

### DIFF
--- a/block_linux.go
+++ b/block_linux.go
@@ -361,6 +361,10 @@ func (ctx *context) disks() []*Disk {
 			driveType = DRIVE_TYPE_ODD
 			busType = BUS_TYPE_SCSI
 			storageController = STORAGE_CONTROLLER_SCSI
+		} else if strings.HasPrefix(dname, "xvd") {
+			driveType = DRIVE_TYPE_HDD
+			busType = BUS_TYPE_SCSI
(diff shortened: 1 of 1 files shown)


retr0h:
Do tests need updated to reflect the change?

jaypipes:
> Do tests need updated to reflect the change?

It would be nice to add a unit test for this, yes :)

That said, I unfortunately don't have any unit tests for this particular function in https://github.com/jaypipes/ghw/blob/93d787280a75b781b3fe1b1cce0e49ae1a813ca0/block_linux_test.go. So, I won't hold up this PR for new tests.

@retr0h do you fancy adding some unit tests for this?

block_linux.go
```diff
@@ -361,6 +361,10 @@ func (ctx *context) disks() []*Disk {
 			driveType = DRIVE_TYPE_ODD
 			busType = BUS_TYPE_SCSI
 			storageController = STORAGE_CONTROLLER_SCSI
+		} else if strings.HasPrefix(dname, "xvd") {
```
jaypipes: @retr0h @mvisonneau if you'd like, please consider adding a code comment here that explains this block is for AWS block device naming and/or link to a URL that explains the AWS naming convention.

Issue #139, opened 2019-07-11
Partitions for MMC/SD disks have no size


jaypipes:
See https://github.com/jaypipes/ghw/issues/132#issuecomment-509151067

The reason for this is because the `partitionSizeBytes()` function is not properly parsing the block device prefix for the partition:

https://github.com/jaypipes/ghw/blob/29869ac89830376f4997b823e455dac004acce4e/block_linux.go#L447-L465

when the disk is MMC/SD.

jaypipes:
An appropriate fix for this should pass the disk name to the `partitionSizeBytes()` function, pre-parsed, instead of passing the partition name and trying to figure out the disk from the partition name.

PR #144, opened 2019-12-03
Fix partitionSizeBytes for MMC/SD block devices

jaypipes:
Due to the way the `ctx.partitionSizeBytes` method was attempting to
determine the disk name by examining the partition name, we were not
properly handling MMC/SD drives. This patch resolves this issue by
moving the regex handling for determining disk names from a single
partition name string into the (deprecated in 0.2)
`ghw.PartitionSizeBytes` function and doing some string prefix matching
that handles all known disk types, including MMC/SD. This is no longer
necessary for the `ctx.partitionSizeBytes` method since this method now
takes the disk name and partition name as separate arguments.

Closes Issue #139

### DIFF
--- a/block_linux.go
+++ b/block_linux.go
@@ -20,9 +20,6 @@ const (
 	sectorSize = 512
 )
 
-var regexNVMeDev = regexp.MustCompile(`^nvme\d+n\d+$`)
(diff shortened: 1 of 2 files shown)


jaypipes:
@LukasMarckmiller, this should fix the partition size in bytes problem for MMC/SD cards. Sorry for the delay!

Issue #145, opened 2019-12-18
compiled failed in macOS


LinkinStars:
```
# github.com/jaypipes/ghw
../../github.com/jaypipes/ghw/dmi_stub.go:15:39: undefined: DMIInfo

Compilation finished with exit code 2
```

you set `// +build !linux` in `dmi_stub.go`, so it can build success in linux.

jaypipes:
thanks for the bug report @LinkinStars! Should be fixed with #147.

PR #146, opened 2019-12-18
Removing leftover dmi_stub and dmi_linux files

jaypipes:
The original dmi.go stuff was split out into disparate chassis,
baseboard, product and bios files, but I forgot to remove the dmi_linux
and dmi_stub files when doing so, which breaks builds on MacOSX.

Closes Issue #145


jaypipes:
oof.

PR #155, opened 2020-02-06
Windows support: cpu, memory

KingRial:
Added partial windows support to cpu and memory

Issue #8

### DIFF
--- a/cpu_windows.go
+++ b/cpu_windows.go
@@ -1,0 +1,55 @@
+// +build !linux
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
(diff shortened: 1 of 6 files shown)


cpu_windows.go
```diff
@@ -0,0 +1,55 @@
...
+			Id:         index, // TODO: how to get a decent "Physical ID" to use ?
+			Model:      description.Name,
+			Vendor:     description.Manufacturer,
+			NumCores:   description.NumberOfCores,
+			NumThreads: description.NumberOfLogicalProcessors,
+		}
+		procs = append(procs, p)
+	}
+	return procs
+}
```
jaypipes: very nice!

PR #156, opened 2020-02-06
Windows support: block

KingRial:
Added partial windows support to block

Issue #8

### DIFF
--- a/block_windows.go
+++ b/block_windows.go
@@ -1,0 +1,233 @@
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
+package ghw
+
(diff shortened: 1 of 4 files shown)


block_windows.go
```diff
@@ -0,0 +1,233 @@
...
+								Type:       diskpartition.Type,
+								IsReadOnly: toReadOnly(diskpartition.Access), // TODO: add information
+							}
+							disk.Partitions = append(disk.Partitions, p)
+							break
+						}
+					}
+				}
+			}
+		}
```
jaypipes: I have an idea on how we can remove the above Go code by just using a JOIN in the WQL statement... but we can work on that together in a followup patch :)

PR #157, opened 2020-02-06
Windows support: net

KingRial:
Added partial windows support to net

Issue #8

### DIFF
--- a/net_windows.go
+++ b/net_windows.go
@@ -11,7 +11,7 @@ import (
 	"github.com/StackExchange/wmi"
 )
 
-const wqlProduct = "SELECT Description, DeviceID, Index, InterfaceIndex, MACAddress, Manufacturer, Name, NetConnectionID, ProductName, ServiceName  FROM Win32_NetworkAdapter"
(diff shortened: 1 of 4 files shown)


net_windows.go
```diff
@@ -0,0 +1,65 @@
...
+		nics = append(nics, nic)
+	}
+
+	return nics
+}
+
+func (ctx *context) netDeviceName(description win32NetworkAdapter) string {
+	var name string
+	if strings.TrimSpace(description.NetConnectionID) != "" {
+		name = description.NetConnectionID + " - " + description.Description
```
jaypipes: I think it would probably be better to just use `NetConnectionID` here and not append the `Description` field if `NetConnectionID` is non-null. Otherwise it's likely you'll end up with NICs named "Wired Connection - Wired Connection" or something like that..

KingRial: I was at the same idea at first but during some tests on the field I found out that their union was more usefull (and sometimes, even mandatory).

Here a quick example on Windows Machine using the command **"go run .\cmd\ghwc\main.go -f yaml net"**:

image

As you can see, their union often deepens the description of the device.

The Latins said: "[Melius abundare quam deficere](https://en.wikipedia.org/wiki/List_of_Latin_phrases_(M)#Melius_abundare)"
We could leave their union in the "name" field or 
hypothesize a new field.

jaypipes: :) OK, you'd convinced me, Riccardo!

PR #163, opened 2020-02-08
Feature: Installed software/packages (Windows only)

KingRial:
This feature allow to collect a list of installed softwares and their installation date.

It's not HW related but the library could be very interesting in the IT environment on assets management,

Example:
image


KingRial:
No problem at all ^^

Feel free to choose how your fantastic library should improve.

PR #162, opened 2020-02-08
Feature: Memory Banks (Windows Only)

KingRial:
When collecting HW informations about memory, it's also very usefull to know the physical memory banks and their specs

Quick example:
image

PR #171, opened 2020-03-10
context.memTotalPhysicalBytes(): Use a Syscall

alrs:
This looks to be either:

1. Me not understanding a corner case that requires getting memory information out of logfiles.

2. A good place to use `syscall`.

If you're open to using system calls instead of parsing files, there are probably a few more commits like this in me.

### DIFF
--- a/memory_linux.go
+++ b/memory_linux.go
@@ -7,31 +7,12 @@ package ghw
 
 import (
 	"bufio"
-	"compress/gzip"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
-	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
-)
-
-const (
-	_WARN_CANNOT_DETERMINE_PHYSICAL_MEMORY = `
-Could not determine total physical bytes of memory. This may
-be due to the host being a virtual machine or container with no
-/var/log/syslog file, or the current user may not have necessary
-privileges to read the syslog. We are falling back to setting the
-total physical amount of memory to the total usable amount of memory
-`
-)
-
-var (
-	// System log lines will look similar to the following:
-	// ... kernel: [0.000000] Memory: 24633272K/25155024K ...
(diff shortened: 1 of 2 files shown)


alrs:
Thinking about this, it's more likely you're looking for something like this, right? 

https://github.com/digitalocean/go-smbios/blob/master/cmd/lsdimms/main.go

jaypipes:
> Thinking about this, it's more likely you're looking for something like this, right?
> 
> https://github.com/digitalocean/go-smbios/blob/master/cmd/lsdimms/main.go

Interesting code there, thanks for linking it here! I think the above may be a good approach for adding the Linux implementation of @KingRial's Windows-specific patch that adds memory-module/DIMM information: https://github.com/jaypipes/ghw/pull/162

Thoughts?

memory_linux.go
```diff
@@ -40,70 +21,25 @@ func (ctx *context) memFillInfo(info *MemoryInfo) error {
...
-					return size
-				}
-			}
-		}
-	}
-	return -1
+func (ctx *context) memTotalPhysicalBytes() (uint64, error) {
+	var info syscall.Sysinfo_t
+	err := syscall.Sysinfo(&info)
+	return info.Totalram, err
```
jaypipes: actually, `sysinfo_t.totalram` is the usable memory amount, not the physical memory amount :)

http://man7.org/linux/man-pages/man2/sysinfo.2.html

The results of sysinfo(2) are replicated in `/proc/meminfo`, which is what is used to get the total usable bytes below. See the comment on (original) line 124 below that explains why we look up the physical memory size using syslog.

That said, I think using the DMI approach (that the digitalocean library uses) would allow us to simply add up the physical size of all DIMMs attached to the host. The only problem with that is, of course, DMI access is a privileged operation so we're back in the same situation as before that caused me to originally write this code... (accessing syslog sometimes requires lower privileges than accessing DMI information, thus the fallback to using syslog instead of DMI)

See the note

PR #198, opened 2020-09-11
Determine total physical memory from sysfs for linux as primary method

cclerget:
This PR allows to first determine on Linux systems the total of physical memory from sysfs (`/sys/devices/system/memory`) as primary method and fallback to syslog file method if `/sys/devices/system/memory` is not present for some reasons

### DIFF
--- a/pkg/memory/memory_linux.go
+++ b/pkg/memory/memory_linux.go
@@ -26,9 +26,10 @@ const (
 	_WARN_CANNOT_DETERMINE_PHYSICAL_MEMORY = `
 Could not determine total physical bytes of memory. This may
 be due to the host being a virtual machine or container with no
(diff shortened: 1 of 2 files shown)


pkg/memory/memory_linux.go
```diff
@@ -55,7 +56,55 @@ func (i *Info) load() error {
...
+			return -1
+		}
+		if strings.TrimSpace(string(s)) != "online" {
+			continue
+		}
+		total += int64(blockSizeBytes)
+	}
+
+	return total
+}
```
jaypipes, line 105: Very, very nice. :+1:

jaypipes (approved):
@cclerget this is truly awesome. thank you!

Issue #66, opened 2018-09-28
Consider checking in a static fs hierarchy for tests


jpo-joyent:
Currently, `go test` runs against your local machine, and some tests fail depending on what hardware that machine actually has.

It might be a good idea to check in a single fs hierarchy to run tests against (using GHW_CHROOT), and could also serve as a handy reference of file formats for contributors without access to various hardware.

jaypipes:
I think this is an excellent idea, @jpo-joyent. I'll try to modify our test setup accordingly and make use of the `GHW_CHROOT` functionality to place a temporary filesystem down for the tests to use.

fromanirh:
Hi there! I'd love to contribute because I'm interested in testing the syfs reading portion of `pkg/cpu`, `pkg/memory` and few others.
Few questions (been doing similar things in the past)
- should the testing fs hierarchy be static (e.g. subtree somewhere under test you just check in the source tree)?
- same as above, but maybe with some .tgz file we uncompress on the fly to run tests?
- should the testing fs hierarchy generated programmatically, so it does exist only when test runs?

jaypipes:
Hi @fromanirh! Check out the `ghw-snapshot` tool:

https://github.com/jaypipes/ghw/blob/master/cmd/ghw-snapshot/main.go

You can run it like so:

```
go run ./cmd/ghw-snapshot/main.go
```

After running that, you will see a `$OS-$ARCH-$HASHSYSTEMNAME.tar.gz` file in the current directory.

You can investigate this snapshot of tmpfs using the `hack/run-against-snapshot.sh` script:

https://github.com/jaypipes/ghw/blob/master/hack/run-against-snapshot.sh

Of course, I'm very happy to accept any and all contributions that improve the test coverage in `ghw`, so please don't let this discourage you! :)

fromanirh:
Thanks for the information @jaypipes ! Far from discouraging, this is a great building block!
I'd like to make use of snapshots systematically in the test suite. I'm most interested in the cpu + topology packages, but we can iterate from there.

Do you think a flow like this could be a good direction?
1. file a PR with few snapshots to be used as test data. Example: few cores (laptop), many cores (12-16 fromm beefy workstation), any-cores multi-numa (from server box)
2. enhance the testsuite to (optionally?) unpack the snapshots before the start up
3. pass the locations to unit tests (env vars?), so they can chroot into them and be fully independent from the host
4. clean up on exit

[...]

PR #203, opened 2020-12-17
ghw-snapshot: capture cpu, memory, topology data

fromanirh:
Needed for https://github.com/jaypipes/ghw/issues/66 (see https://github.com/jaypipes/ghw/issues/66#issuecomment-738618640 and following)

Add support to capture data ghw cares about in
`/sys/system/devices/{cpu,node}.`
With this patch, we can create snapshot we can reuse
to test the `cpu`, `memory` and `topology` packages using
the codeto consume snapshot added in commit 1a9fae7

There is quite a lot of opportunity to extract code
from ghw-snapshot and add it to the newly created `snapshot`
pkg, to make it easier to reuse and more testable.

This refactoring is deferred to a later change because
the new logic added is complex enough to stand in its own.


fromanirh:
it seems `ghw-snapshot` is saving absolute (not relative) links for block devices pointing to the original scratchdir. This breaks block devices snapshot usage in my case.
Working on a fix, can be squashed in this PR or an additional one. Any preferences?
Same goes for snapshots: I'm collecting the promised reference snapshot from an handful systems I have access to, I can either push a followup PR or add here in the next few days. Of course they're going to be a bunch of blobs from git perspective.
I'd be happy to do whatever works best forreviewers!

jaypipes:
> it seems `ghw-snapshot` is saving absolute (not relative) links for block devices pointing to the original scratchdir. This breaks block devices snapshot usage in my case.

Yes, I did that on purpose after failing miserably with relative links :) The reason is because if you pass a directory containing relative links to Docker as a volume mount, Docker won't read the contents of the relative link (for security reasons). Docker considers a relative link to be simply a file that contains a string with a filepath in it:

https://github.com/moby/moby/issues/1676

> Working on a fix, can be squashed in this PR or an additional one. Any preferences?

I'd definitely be interested in your solution to this :) Please file a separate PR, though, as it will help for reviewing!

Best,
-jay

fromanirh:
> > it seems `ghw-snapshot` is saving absolute (not relative) links for block devices pointing to the original scratchdir. This breaks block devices snapshot usage in my case.
> 
> Yes, I did that on purpose after failing miserably with relative links :) The reason is because if you pass a directory containing relative links to Docker as a volume mount, Docker won't read the contents of the relative link (for security reasons). Docker considers a relative link to be simply a file that contains a string with a filepath in it:
> 
> [moby/moby#1676](https://github.com/moby/moby/issues/1676)\

Ah, darn, I added to this PR because I'm using the same approach I followed for `/sys/devices/system/node/node0/cpu?` (which symlink back to `/sys/devices/system/cpu/cpu?` :( I'll test with docker (and podman) then.

>  > Working on a fix, can be squashed in this PR or an additional one. Any preferences?
> 
> I'd definitely be interested in your solution to this :) Please file a separate PR, though, as it will help for reviewing!

OK, so I already made a little mess so I'll stay put and wait for further instructions. Apologies!

fromanirh:
I think we have a good resolution onhttps://github.com/jaypipes/ghw/pull/205 - I will rebase this on top of it when we reach consensus.

fromanirh:
works, but forgot to snapshot the relevant content under `/sys/devices/system/memory`. Will update ASAP.

fromanirh:
updated! now the snapshot captures the `/sys/devices/system/memory` ghw cares about. Please note the example snapshot was removed from *this* PR. I'll add again shortly (in another PR) once I get access again to my test system.

jaypipes (approved):
++

PR #209, opened 2021-01-12
add PCI device info to ghwc

jaypipes:
Adds a new `ghwc pci` subcommand that lists PCI devices. Deprecates the
old PCIInfo.Products, PCIInfo.Classes and PCIInfo.Vendors maps because
these are now fully contained in the github.com/jaypipes/pcidb library.

```
[jaypipes@thelio ghw]$ go run cmd/ghwc/main.go pci
PCI (45 devices)
[jaypipes@thelio ghw]$ go run cmd/ghwc/main.go pci --format yaml | head -n30
pci:
  Devices:
  - address: "0000:00:00.0"
    class:
      id: "06"
      name: Bridge
    product:
      id: "1450"
      name: Family 17h (Models 00h-0fh) Root Complex
    programming_interface:
      id: "00"
      name: unknown
    subclass:
      id: "00"
      name: Host bridge
    subsystem:
      id: "1450"
      name: unknown
    vendor:
      id: "1022"
      name: Advanced Micro Devices, Inc. [AMD]
  - address: "0000:00:00.2"
    class:
      id: "08"
      name: Generic system peripheral
    product:
      id: "1451"
      name: Family 17h (Models 00h-0fh) I/O Memory Management Unit
    programming_interface:
      id: "00"
```

Issue #207

PR #210, opened 2021-01-15
pci: JSON device marshalling fixes

fromanirh:
PR https://github.com/jaypipes/ghw/pull/209 exposed a mashalling issue with some devices.
The issue is the quoting of " from the description of some devices (e.g. 2.5" device).

Add a test to demonstrate the bug and a possible fix.

### DIFF
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -7,7 +7,7 @@
 package pci
 
 import (
-	"bytes"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


fromanirh:
I added `pci.ParseDevice` because it seems a good compromise and an acceptable addition to the API; I can add a ghw snapshot which highlights the issue and which allows us to demonstrate the bug without API changes, but it will take me some more time.

pkg/pci/pci.go
```diff
@@ -40,69 +40,54 @@ type Device struct {
...
 
+type devIdent struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type devMarshallable struct {
+	Address   string   `json:"address"`
+	Vendor    devIdent `json:"vendor"`
+	Product   devIdent `json:"vendor"`
```
jaypipes, line 51: copy/paste mistake ... the json tag should be "product" :)

pkg/pci/pci.go
```diff
@@ -40,69 +40,54 @@ type Device struct {
...
+	Subsystem devIdent `json:"vendor"`
```
jaypipes, line 52: and this should be "subsystem"

pkg/pci/pci_linux.go
```diff
@@ -233,7 +238,21 @@ func (info *Info) GetDevice(address string) *Device {
 	if modaliasInfo == nil {
 		return nil
 	}
+	return info.getDeviceFromModaliasInfo(address, modaliasInfo)
+}
+
+// ParseDevice returns a pointer to a Device given its describing data.
+// The PCI device obtained this way may not exist in the system;
+// use GetDevice to get a *Device which is found in the system
+func (info *Info) ParseDevice(address, modalias string) *Device {
```
jaypipes, line 247: technically this is introducing a public exported interface, and I realize you did this in order to unit-test it in the pci_test.go file which imports the `pkg/pci` package in order to test the exported interfaces. So, I'm OK with this.

jaypipes (changes requested):
Very nice work @fromanirh! Just a little copy/pasta mistake to fix up otherwise, lgtm.

jaypipes:
> I added `pci.ParseDevice` because it seems a good compromise and an acceptable addition to the API (ok, ok, disclosure: it's the lazy way!); I can add a ghw snapshot which highlights the issue and which allows us to demonstrate the bug without API changes, but it will take me some more time.

Yup, I'm cool with it :)

jaypipes (approved):
\o/ thanks @fromanirh!

Issue #175, opened 2020-03-17
Provide `ghw.WithDisableWarnings()` function


tredoe:
https://github.com/jaypipes/ghw#disabling-warning-messages

Now, there is to use an environment variable to disable the warnings, but it is not an option when the package is going to be compiled into a main program.

The best would be to add a global variable at 'ghw' to disable them.

jaypipes:
Are you referring to the `ghwc` CLI tool? Or are you referring to `import`ing the ghw library? In either case, the way to disable warning messages is to set the environment variable that the library looks for, which is `GHW_DISABLE_WARNINGS`.

If you are `import`ing the ghw library into your own executable and wish to always disable these warnings, then *your* code should simply call:

```go
import os

os.Setenv("GHW_DISABLE_WARNINGS", "1")
```

https://golang.org/pkg/os/#Setenv

before calling any `ghw` library functions.

Make sense?

tredoe:
It does not make sense that a library use environment variables to enable/disable features on it.

jaypipes:
@tredoe I could add a `ghw.WithDisableWarnings()` technique:

```go
net, err := ghw.Network(ghw.WithDisableWarnings())
```

similar to the `ghw.WithChroot()` function:

https://github.com/jaypipes/ghw#overriding-the-root-mountpoint-ghw-uses

https://github.com/jaypipes/ghw/blob/14923eaaa4d8378051f741694e261747d16fc05e/options.go#L34

that can be used to change the root directory path in addition to the `GHW_CHROOT` environment variable.

Would that be more amenable to you?

tredoe:
Oh yes, please

jaypipes:
> It does not make sense that a library use environment variables to enable/disable features on it.

Plenty of libraries control features using environment variables, by the way.

Examples:

* **Logrus**: https://github.com/sirupsen/logrus/blob/53000c4c0f981521df1bada12a9e9ea20728fb62/text_formatter.go#L116
* **Ginkgo**: https://github.com/onsi/ginkgo/blob/6a07ea2adeafc7c9dc2a9b38ffc4154f9159f467/ginkgo/run_command.go#L130
* **aws-sdk-go**: https://github.com/aws/aws-sdk-go/blob/bef02444773a49eaf30cdd615920b56896827c06/aws/credentials/env_provider.go#L44
[...]

PR #216, opened 2021-01-25
warning: support new option to override default

fromanirh:
This PR wants to solve https://github.com/jaypipes/ghw/issues/175

This patch add support for a new option, which allow
the caller code to pass a new Alerter to ghw.
`Alerter` is a dead-simple interface, supporting only a
`Printf` method much like the stdlib `log` package.
ghw will send all the warnings to this method.

We use a so simple interface because we want to generalize a bit
the existing `util.Warn` concept, but is not clear if ghw will benefit
of a more complete logging interface; should that be the case, we
can build on the concept we laid in this patch.

Client code can programmatically override the Alerter, or
users can using the faimiliar `GHW_DISABLE_WARNINGS` environ variable,
which keep functioning as usual.

ghw can be configured to swallow all the warnings or to keep emitting them
on stderr, like it used to do up until this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,20 @@ $ GHW_DISABLE_WARNINGS=1 ghwc memory
 memory (24GB physical, 24GB usable)
 ```
 
+You can disable warning programmatically using the `WithAlerter` option:
+
+```go
+
+import (
+	"io/ioutil"
+	"log"
+
+	"github.com/jaypipes/ghw"
+)
+
+mem, err := ghw.Memory(ghw.WithAlerter(log.New(ioutil.Discard, "", 0))
+```
+
 ### Memory
 
 Information about the host computer's memory can be retrieved using the
--- a/README.md
+++ b/README.md
@@ -207,18 +207,19 @@ $ GHW_DISABLE_WARNINGS=1 ghwc memory
 memory (24GB physical, 24GB usable)
 ```
 
-You can disable warning programmatically using the `WithAlerter` option:
+You can disable warning programmatically using the `WithNullAlerter` option:
 
 ```go
 
 import (
-	"io/ioutil"
-	"log"
-
 	"github.com/jaypipes/ghw"
 )
 
-mem, err := ghw.Memory(ghw.WithAlerter(log.New(ioutil.Discard, "", 0))
+mem, err := ghw.Memory(ghw.WithNullAlerter())
+
+You may also supply an `Alerter` to ghw to handle the warnings.
+In this case, please check the `option.Alerter` interface (compatible with stdlib's
+log.Logger type) and the `ghw.WithAlerter()` function
 ```
 
 ### Memory
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ $ GHW_DISABLE_WARNINGS=1 ghwc memory
 memory (24GB physical, 24GB usable)
 ```
 
-You can disable warning programmatically using the `WithNullAlerter` option:
+You can disable warning programmatically using the `WithDisableWarnings` option:
 
 ```go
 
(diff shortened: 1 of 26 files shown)


fromanirh:
WIP because while the code is {mostly, hopefully} straightforward, there are quite some design concepts to properly evaluate here.
For example: do default (trivial) Alerters belong in the `option` pkg?

Also, tests are completely missing and we need at least basic coverage.

pkg/context/context.go
```diff
@@ -18,12 +18,15 @@ type Context struct {
 	SnapshotPath      string
 	SnapshotRoot      string
 	SnapshotExclusive bool
+	SnapshotPreserve  bool
```
fromanirh, line 21: bad rebase, will fix in the next upload which I need to do anyway to add tests.

jaypipes (commented):
:+1: I like this.

pkg/context/context.go
jaypipes: @fromanirh you still working on this patch, right?

fromanirh: yes! totally ongoing, but not sure I can make it for end of the week (aka 0.7 cutoff).

fromanirh:
Ok, managed to do some of the updates I planned and I'm reasonnably happy with the state of the PR. We can iterate and improve a bit later; WIP removed, PR fully reviewable!

fromanirh:
fixed stale commit message. No code changes.

README.md
```diff
@@ -207,6 +207,20 @@ $ GHW_DISABLE_WARNINGS=1 ghwc memory
...
+```go
+
+import (
+	"io/ioutil"
+	"log"
+
+	"github.com/jaypipes/ghw"
+)
+
+mem, err := ghw.Memory(ghw.WithAlerter(log.New(ioutil.Discard, "", 0))
```
jaypipes, line 221: we could make a singleton `ghw.NullAlerter` to make this user experience a bit better (and avoid the user needing to import `io/ioutil" and "log" packages themselves.

For instance, in `pkg/option/option.go` we could have:

```go
import (
    "io/ioutil"
     "log"
)

var (
    NullAlerter = &Alerter{
        log.New(ioutil.Discard, "", 0),
    }
)
```

and then the user would just need to do:

```go
mem, err := ghw.Memory(ghw.WithAlerter(ghw.NullAlerter))
```

We could make it even easier by making a `ghw.WithNullAlerter()` function, too, in `pkg/option/option.go`:

```go
// WithNullAlerter sets No-op alerting options for ghw
func WithNullAlerter(alerter Alerter) *Option {
	return &Option{
		Alerter: NullAlerter,
	}
}
```

after which all the user would need to do would be this:


```go
mem, err := ghw.Memory(ghw.WithNullAlerter())
```

jaypipes (commented):
suggestion inline :)

README.md
fromanirh: I like it, will update the PR soon(ish) to add this change.

README.md
```diff
@@ -207,18 +207,19 @@ $ GHW_DISABLE_WARNINGS=1 ghwc memory
...
 	"github.com/jaypipes/ghw"
 )
 
-mem, err := ghw.Memory(ghw.WithAlerter(log.New(ioutil.Discard, "", 0))
+mem, err := ghw.Memory(ghw.WithNullAlerter())
+
+You may also supply an `Alerter` to ghw to handle the warnings.
+In this case, please check the `option.Alerter` interface (compatible with stdlib's
+log.Logger type) and the `ghw.WithAlerter()` function
 ```
```
jaypipes, line 223: :+1:

jaypipes (approved):
eccellente! grazie @fromanirh! :)

fromanirh:
just wild thought reviewing https://github.com/jaypipes/ghw/issues/175: do we want to add an alias, perhaps in another PR, to minimize the surprise for users? Programmatically we have `WithNullAlerter`, using env vars we have `GHW_DISABLE_WARNING`.
We can add
```
var WithDisableWarnings = option.WithNullAlerter
```

I believe we should keep WithNullAlerter because it's more powerful construct.

fromanirh:
Done! Now I think we have everything we need to call https://github.com/jaypipes/ghw/issues/175 really done.

jaypipes (approved):
Excellent, thank you @fromanirh!

Issue #213, opened 2021-01-21
Fall back on ID_SERIAL if ID_SERIAL_SHORT is missing


nevermore-muyi:
If I cannot get the ID_SERIAL_SHORT, we may should get ID_SERIAL for serial number?

jaypipes:
Many systems require root privileges to access chassis and baseboard serial information. Please make sure you check for WARNING messages from ghw like this:

```
WARNING: Unable to read chassis_serial: open /sys/class/dmi/id/chassis_serial: permission denied
```

If you run a tool like `ghwc` as `sudo` and continue to not see serial information, please let me know.

nevermore-muyi:
@jaypipes I used root to use ghwc, but there is not a  ID_SERIAL_SHORT ,the output like this:

> E:ID_FS_TYPE=LVM2_member
> E:ID_FS_USAGE=raid
> E:ID_FS_UUID=sTW2DA-05SV-FWe4-XmY5-jKL2-YgRe-soUYjw
> E:ID_FS_UUID_ENC=sTW2DA-05SV-FWe4-XmY5-jKL2-YgRe-soUYjw
> E:ID_FS_VERSION=LVM2 001
> E:ID_PATH=pci-0000:02:02.0
> E:ID_PATH_TAG=pci-0000_02_02_0
> E:ID_SERIAL=37044b0a-e68a-4e83-a

jaypipes:
Ah, I understand you now. OK, the code I need to change is here:

https://github.com/jaypipes/ghw/blob/9584f3470707b0076dda2550a2ad1da28bd2d383/pkg/block/block_linux.go#L136-L141

I'll push a fix shortly to fallback on ID_SERIAL if ID_SERIAL_SHORT is missing.

nevermore-muyi:
And when will the next version can be published?

jaypipes:
> And when will the next version can be published?

I'm aiming for Friday this week (or before). I have a dependency on the pcidb library that I need to release first.

nevermore-muyi:
I am looking forward to the next version.

jaypipes:
> I am looking forward to the next version.

@nevermore-muyi: it's out :) https://github.com/jaypipes/ghw/releases/tag/v0.7.0

nevermore-muyi:
> > And when will the next version can be published?
> 
> I'm aiming for Friday this week (or before). I have a dependency on the pcidb library that I need to release first.

OK, thanks, i am going to try it.

PR #217, opened 2021-01-25
fall back to ID_SERIAL if ID_SERIAL_SHORT missing

jaypipes:
Some systems don't have an ID_SERIAL_SHORT value. In these cases,
instead of returning UNKNOWN, let's fall back to ID_SERIAL.

Issue #213


fromanirh:
lgtm!

PR #219, opened 2021-01-25
PCI: Add device revision number

pearsonk:
Providing access to the revision number of hardware can be useful in
identifying or diagnosing problems introduced across versions of single
product line.

Signed-off-by: Kevin Pearson <kevin.pearson@ortmanconsulting.com>

### DIFF
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -44,6 +44,28 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
 	)
 }
 
+func getDeviceRevision(ctx *context.Context, address string) string {
+	paths := linuxpath.New(ctx)
+	pciAddr := AddressFromString(address)
+	if pciAddr == nil {
+		return ""
+	}
+	revisionPath := filepath.Join(
+		paths.SysBusPciDevices,
+		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Slot+"."+pciAddr.Function,
+		"revision",
+	)
+
+	if _, err := os.Stat(revisionPath); err != nil {
+		return ""
+	}
+	revision, err := ioutil.ReadFile(revisionPath)
+	if err != nil {
+		return ""
+	}
+	return string(revision)
+}
+
 type deviceModaliasInfo struct {
 	vendorID     string
 	productID    string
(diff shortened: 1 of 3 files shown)


pkg/pci/pci_linux.go
```diff
@@ -44,6 +44,28 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
...
+	}
+	revisionPath := filepath.Join(
+		paths.SysBusPciDevices,
+		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Slot+"."+pciAddr.Function,
+		"revision",
+	)
+
+	if _, err := os.Stat(revisionPath); err != nil {
+		return ""
+	}
```
fromanirh, line 61: this is not wrong, but it seems we can just call ReadFile directly instead of checking with `os.Stat` before.

pkg/pci/pci_linux.go
```diff
@@ -44,6 +44,28 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
...
 
+func getDeviceRevision(ctx *context.Context, address string) string {
+	paths := linuxpath.New(ctx)
+	pciAddr := AddressFromString(address)
+	if pciAddr == nil {
+		return ""
+	}
+	revisionPath := filepath.Join(
+		paths.SysBusPciDevices,
+		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Slot+"."+pciAddr.Function,
```
fromanirh, line 55: this suggests we may want to add a `String` method to `pci.Address` (not in this PR)

pkg/pci/pci_linux.go
pearsonk: I concur. I implemented it in this fashion to mirror the existing approach for reading the modalias file, for the sake of consistency. I'm not against the simplified approach of attempting a read and reacting on the returned error

fromanirh: I don't have strong opinions here. Both approaches have their merits.

fromanirh:
/lgtm
just for the record to reach closure on my side :)

pkg/pci/pci_linux.go
jaypipes: yeah, totally fine to use existing pattern here.

jaypipes (approved):
Thanks @pearsonk! Great stuff :)

Issue #92, opened 2018-11-19
Grab information on VFs for SR-IOV PFs


jaypipes:
It's possible to retrieve the number of virtual functions (VFs) on an SR-IOV physical function (PF) on Linux by querying the following pseudo-files:

 * `/sys/class/net/$NIC_NAME/device/sriov_numvfs`: The number of active VFs for the PF
* `/sys/class/net/$NIC_NAME/device/sriov_totalvfs`: The maximum number of VFs the PF can accomodate

fromanirh:
Hi there, I can contribute here, depending on the decision we take in https://github.com/jaypipes/ghw/issues/66 (and also https://github.com/jaypipes/ghw/issues/201 :) )

fromanirh:
I'd totally happy to work in this area. I have access to boxes with low and high end SRIOV cards and I can provide snapshots (albeit most likely for low-end cards only) to enable others.

jaypipes:
> I have access to boxes with low and high end SRIOV cards and I can provide snapshots (albeit most likely for low-end cards only) to enable others.

That is awesome news, @fromanirh! Would be great to get contributions from you in this area!

fromanirh:
first draft, RFC for API: https://github.com/jaypipes/ghw/pull/230

PR #221, opened 2021-02-04
pkg: net: report backing device address

fromanirh:
Report the backing device address - currently PCI only -
in the network interface data.
This allows the client code to fetch the device information using
the specific package, for example `pkg/pci`.

We prefer to add the address, and not a pointer to the relevant
backing device object, the make the net module less coupled with
the other hardware modules.

### DIFF
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -25,8 +25,8 @@ type NIC struct {
 	MacAddress   string           `json:"mac_address"`
 	IsVirtual    bool             `json:"is_virtual"`
 	Capabilities []*NICCapability `json:"capabilities"`
-	// TODO(jaypipes): Add PCI field for accessing PCI device information
-	// PCI *PCIDevice `json:"pci"`
+	PCIAddress   *string          `json:"pci_address,omitempty"`
+	// TODO(fromani): add other hw addresses (USB) when we support them
 }
 
 func (n *NIC) String() string {
--- a/pkg/net/net_linux.go
+++ b/pkg/net/net_linux.go
@@ -67,6 +67,9 @@ func nics(ctx *context.Context) []*NIC {
 		} else {
 			nic.Capabilities = []*NICCapability{}
 		}
+
+		nic.PCIAddress = netDevicePCIAddress(paths.SysClassNet, filename)
+
 		nics = append(nics, nic)
 	}
 	return nics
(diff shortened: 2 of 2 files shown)


fromanirh:
WIP because we need tests; to enable tests, we need to have a ghw snapshot in the source tree; I will provide it with a PR soon enough.

fromanirh:
To move forward, we need first to move forward on https://github.com/jaypipes/ghw/pull/224

fromanirh:
This PR is a precondition to solve https://github.com/jaypipes/ghw/issues/92

pkg/net/net.go
```diff
@@ -20,13 +20,17 @@ type NICCapability struct {
 	CanEnable bool   `json:"can_enable"`
 }
 
+type NICAddress struct {
+	PCI string `json:"pci"`
+	// TODO(fromani): add other hw addresses (USB) when we support them
```
jaypipes, line 25: So this would essentially be a union type, yes? There would be a set of different address fields, and only one would be expected to be set?

pkg/net/net_linux.go
```diff
@@ -163,3 +166,56 @@ func netParseEthtoolFeature(line string) *NICCapability {
...
+	// device path to its full sysfs path.
+	// say we start with netDevDir="/sys/class/net" and netDevName="enp0s31f6"
+	netPath := filepath.Join(netDevDir, netDevName)
+	dest, err := os.Readlink(netPath)
+	if err != nil {
+		// bail out with empty value
+		return nicAddr
+	}
+	// now we have something like dest="../../devices/pci0000:00/0000:00:1f.6/net/enp0s31f6"
+	// remember the path is relative to netDevDir="/sys/class/net"
```
jaypipes, line 187: :+1: the comments in this function are exactly the kind of great code comments that `ghw` promotes: explain the details and the *why*. bravo.

pkg/net/net.go
```diff
@@ -20,13 +20,17 @@ type NICCapability struct {
...
+}
+
 type NIC struct {
 	Name         string           `json:"name"`
 	MacAddress   string           `json:"mac_address"`
 	IsVirtual    bool             `json:"is_virtual"`
 	Capabilities []*NICCapability `json:"capabilities"`
-	// TODO(jaypipes): Add PCI field for accessing PCI device information
-	// PCI *PCIDevice `json:"pci"`
+	HWAddress    NICAddress       `json:"hw_address"`
```
jaypipes, line 33: I think I'd almost prefer to have a `*string` field called `PCIAddress` that is nullable to indicate the field isn't always filled in. If and when in the future we support non-PCI hardware addreses like USB, we can add another `*string` field called `USBAddress`.

pkg/net/net_linux.go
```diff
@@ -163,3 +166,56 @@ func netParseEthtoolFeature(line string) *NICCapability {
 		CanEnable: !fixed,
 	}
 }
+
+func netDeviceHWAddress(netDevDir, netDevName string) NICAddress {
```
jaypipes, line 170: and I think it would be best to just name this function netDevicePCIAddress() since the function only looks up the PCI address (and have it return a `*string`)

jaypipes (changes requested):
@fromanirh thank you for the PR! Left a couple notes and suggestions inline for ya.

pkg/net/net.go
fromanirh: Yes, that's the intent. The reason is I'm aware of pci-connected (desktop, servers...) and usb-connected (laptops, raspberry pi) network interfaces. So that's at least two.
I'm not sure what's the best way to model this facet and I'm very open to suggestions.

pkg/net/net_linux.go
fromanirh: Thanks!

pkg/net/net.go
fromanirh: Works for me. I was never really fond of `hw_address`. It creates some confusion with the NIC physical address (MAC) which I prefer to avoid.

pkg/net/net_linux.go
fromanirh: sure, on it.

jaypipes (approved):
Awesome work, @fromanirh :) Thank you!

jaypipes:
@fromanirh for a future PR, feel free to update the README.md to list this new PCIAddress field.

fromanirh:
> @fromanirh for a future PR, feel free to update the README.md to list this new PCIAddress field.

Sure thing! https://github.com/jaypipes/ghw/pull/226 (actually I should have done this earlier :\ )

PR #223, opened 2021-02-04
pci: add String() method to pci.Device

fromanirh:
We have AddressFromString(), so it seems natural to add
the symmetric, which we can also use in few places internally
to make the code more readable.


jaypipes (approved):
Nice :)

PR #224, opened 2021-02-05
snapshot: gather full pci tree, handling bridges

fromanirh:
the current implementation of the PCI device snapshot only scans the first level in the PCI tree (actually, PCI forest). Anything attached to bridges is ignored. This was unintended, and this PR aims to fix it acknowledging bridges explictely and handling them correctly. In practices, this means to scan also the devices attached to each bridge, recursively.

Note this affects only the snapshotting code.

### DIFF
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/jaypipes/ghw/pkg/context"
 	"github.com/jaypipes/ghw/pkg/linuxpath"
+	pciaddr "github.com/jaypipes/ghw/pkg/pci/address"
 	"github.com/jaypipes/ghw/pkg/topology"
 	"github.com/jaypipes/ghw/pkg/util"
 )
@@ -34,7 +35,7 @@ func (i *Info) load() error {
 
 func getDeviceModaliasPath(ctx *context.Context, address string) string {
 	paths := linuxpath.New(ctx)
-	pciAddr := AddressFromString(address)
+	pciAddr := pciaddr.FromString(address)
 	if pciAddr == nil {
 		return ""
 	}
@@ -47,7 +48,7 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
 
 func getDeviceRevision(ctx *context.Context, address string) string {
 	paths := linuxpath.New(ctx)
-	pciAddr := AddressFromString(address)
+	pciAddr := pciaddr.FromString(address)
 	if pciAddr == nil {
 		return ""
 	}
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -6,7 +6,6 @@
 package pci
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -277,11 +276,13 @@ func findPCIProgrammingInterface(
 func (info *Info) GetDevice(address string) *Device {
 	fp := getDeviceModaliasPath(info.ctx, address)
 	if fp == "" {
+		info.ctx.Warn("error finding modalias info for device %q", address)
 		return nil
 	}
 
 	modaliasInfo := parseModaliasFile(fp)
 	if modaliasInfo == nil {
+		info.ctx.Warn("error parsing modalias info for device %q", address)
 		return nil
 	}
 
@@ -355,7 +356,7 @@ func (info *Info) ListDevices() []*Device {
 	// address and append to the returned array.
 	links, err := ioutil.ReadDir(paths.SysBusPciDevices)
 	if err != nil {
-		_, _ = fmt.Fprintf(os.Stderr, "error: failed to read /sys/bus/pci/devices")
+		info.ctx.Warn("failed to read /sys/bus/pci/devices")
 		return nil
 	}
 	var dev *Device
@@ -363,7 +364,7 @@ func (info *Info) ListDevices() []*Device {
 		addr := link.Name()
 		dev = info.GetDevice(addr)
 		if dev == nil {
-			_, _ = fmt.Fprintf(os.Stderr, "error: failed to get device information for PCI address %s\n", addr)
+			info.ctx.Warn("failed to get device information for PCI address %s", addr)
 		} else {
 			devs = append(devs, dev)
 		}
--- a/pkg/snapshot/clonetree_net.go
+++ b/pkg/snapshot/clonetree_net.go
@@ -14,6 +14,8 @@ import (
 )
 
 const (
+	// warning: don't use the context package here, this means not even the linuxpath package.
+	// TODO(fromani) remove the path duplication
 	sysClassNet = "/sys/class/net"
 )
 
--- a/pkg/snapshot/clonetree_pci.go
+++ b/pkg/snapshot/clonetree_pci.go
@@ -1,0 +1,148 @@
+//
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
+package snapshot
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	pciaddr "github.com/jaypipes/ghw/pkg/pci/address"
+)
+
+const (
+	// root directory: entry point to start scanning the PCI forest
+	// warning: don't use the context package here, this means not even the linuxpath package.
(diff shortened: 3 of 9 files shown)


fromanirh:
Ready for design review; docstrings are missing, and I need to review unit tests. hence the WIP. Conflicts with https://github.com/jaypipes/ghw/pull/223

pkg/snapshot/clonetree_pci.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	//        subclass ---++
+	// class -----------++||
+	//                  VVVV
+	PCI_BRIDGE_PCI = "0x060400"
+)
+
+// this is a hack to avoid this circolar import:
+// snapshot -> pci -> context -> snapshot
+func isPCIBridge(entryPath string) bool {
```
fromanirh, line 114: not fond of this. I found a way to get rid of this but it will require more internal changes in the `pci` subpkg. Is this an option?
Demo: https://github.com/fromanirh/ghw/commits/snapshot-pci-tree-device

pkg/pci/pci_linux.go
```diff
@@ -16,9 +16,16 @@ import (
...
 	"github.com/jaypipes/ghw/pkg/linuxpath"
+	pciaddr "github.com/jaypipes/ghw/pkg/pci/address"
 	"github.com/jaypipes/ghw/pkg/util"
 )
 
+// backward compatibility, to be removed in 1.0.0
+type Address pciaddr.Address
+
+// backward compatibility, to be removed in 1.0.0
+var AddressFromString = pciaddr.FromString
```
fromanirh, line 27: this should be moved in `pci.go`

fromanirh:
clarified (again) the commit message

fromanirh:
still not completely happy. Let me evaluate another approach, will update ASAP.

fromanirh:
I'm still testing https://github.com/jaypipes/ghw/pull/224/commits/9b19ce6d5a9efcb4c1d9416effed4ba131cf7574 and I'm happier with this approach so far.

pkg/snapshot/clonetree_pci.go
fromanirh: not relevant anymore!

pkg/pci/pci_linux.go
fromanirh: no longer relevant

fromanirh:
rebased on top of https://github.com/jaypipes/ghw/pull/227 . I think this PR is now good enough.

fromanirh:
rebased again. No code changes.

pkg/snapshot/clonetree_net.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 const (
+	// warning: don't use the context package here, this means not even the linuxpath package.
+	// TODO(fromani) remove the path duplication
```
jaypipes, line 18: Is the above still true now the you've split pci address into its own package? Why don't we want to use the context package here?

pkg/snapshot/clonetree_pci.go
```diff
@@ -0,0 +1,135 @@
...
+		fileSpecs = append(fileSpecs, specs...)
+	}
+	return fileSpecs
+}
+
+// scanPCIDeviceRoot reports a slice of glob patterns which represent the pseudofiles
+// ghw cares about pertaining to all the PCI devices connected to the bus connected from the
+// given root; usually (but not always) a CPU packages has 1+ PCI(e) roots, forming the first
+// level; more PCI bridges are (usually) attached to this level, creating deep nested trees.
+// hence we need to scan all possible roots, to make sure not to miss important devices.
```
jaypipes, line 49: I'm glad you were able to see the forest for the trees. ;)

pkg/snapshot/clonetree_pci.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return fileSpecs, pciRoots
+}
+
+func findPCIEntryFromPath(root, entryName string) string {
+	entryPath := filepath.Join(root, entryName)
+	fi, err := os.Lstat(entryPath)
+	if err != nil {
+		trace("stat(%s) failed: %v\n", entryPath, err)
+		return entryPath // what else we can do?
```
jaypipes, line 95: we could return err here and above on line 75 we could `continue`... that way we don't bury the error here (even though we're `trace()`ing it.

pkg/snapshot/clonetree_pci.go
```diff
@@ -0,0 +1,135 @@
...
+	if fi.Mode()&os.ModeSymlink == 0 {
+		// regular file, nothing to resolve
+		return entryPath
+	}
+	// resolve symlink
+	target, err := os.Readlink(entryPath)
+	trace("entry %q is symlink resolved to %q\n", entryPath, target)
+	if err != nil {
+		trace("readlink(%s) failed: %v - skipped\n", entryPath, err)
+		return entryPath // what else we can do?
```
jaypipes, line 106: same as above.

pkg/snapshot/clonetree_pci.go
```diff
@@ -0,0 +1,135 @@
...
+		if addr := pciaddr.FromString(subNode.Name()); addr != nil {
+			// we got an entry in the directory pertaining to this device
+			// which is a directory itself and it is named like a PCI address.
+			// Hence we infer the device we are considering is a PCI bridge of sorts.
+			// This is is indeed a bit brutal, but the only possible alternative
+			// (besides blindly copying everything in /sys/bus/pci/devices) is
+			// to detect the type of the device and pick only the bridges.
+			// This approach duplicates the logic with the `pci` subkpg
+			// - or forces us into awkward dep cycles, and has poorer forward
+			// compatibility.
```
jaypipes, line 130: good explanation. :+1:

jaypipes (commented):
Great work, @fromanirh. Just one little request inline to return an error from `findPCIEntryFromPath`, otherwise this is awesome.

pkg/snapshot/clonetree_net.go
fromanirh: thing is: context wants to manage snapshots, to transparently consume them. And this is by design, and we should keep it. So if we import `linuxpath` here... 
```
go vet 
package github.com/jaypipes/ghw
	imports github.com/jaypipes/ghw/pkg/baseboard
	imports github.com/jaypipes/ghw/pkg/context
	imports github.com/jaypipes/ghw/pkg/snapshot
	imports github.com/jaypipes/ghw/pkg/linuxpath
	imports github.com/jaypipes/ghw/pkg/context: import cycle not allowed
```
obviously not very different importing `context`, just for the record:
```
go vet 
package github.com/jaypipes/ghw
	imports github.com/jaypipes/ghw/pkg/baseboard
	imports github.com/jaypipes/ghw/pkg/context
	imports github.com/jaypipes/ghw/pkg/snapshot
	imports github.com/jaypipes/ghw/pkg/context: import cycle not allowed
```
There is however another way out: we can split the `clonetree` code (which needs to walk through the paths) code from the `snapshot` package, or we can split the `snapshot package` in a simpler subpackage which consumes snapshot, and which `context` will import, and in another, more complex and richer package which will create snapshots and which will consume (not be consumer of) `context`.

I don't have strong preferences here and each approach has pros and cons, so, I'd be happy to fo the direction you like most.

pkg/snapshot/clonetree_pci.go
fromanirh: :)

pkg/snapshot/clonetree_pci.go
fromanirh: this is probably a better direction, yes. Let me try & test this out a bit on a couple systems - I want to run against more complex PCI topologies.

pkg/snapshot/clonetree_pci.go
fromanirh: ack

pkg/snapshot/clonetree_pci.go
fromanirh: thanks. The original implementation duplicated code from `pci` package to detect the pci bridges. This is probably more conceptually like the right thing to do, but in practice is less maintenable.

fromanirh: also: we cannot just consume the `pci` package here (hence the `pciaddress` split!) again because of the circular dep to `context`.

jaypipes (approved):
thank you @fromanirh! :)

PR #233, opened 2021-03-09
janitorial: modernize error detection

fromanirh:
Reading the documentation of `os.IsNotExist` and similar functions
(see https://golang.org/pkg/os/#IsNotExist), we learn that

```
This function predates errors.Is. It only supports errors returned
by the os package. New code should use errors.Is(err, os.ErrNotExist).
```

ghw uses these functions but sparsely, so we take the chance
in this PR to modernize the codebase.

No intended changes in behaviour.

### DIFF
--- a/pkg/gpu/gpu_test.go
+++ b/pkg/gpu/gpu_test.go
@@ -7,6 +7,7 @@
 package gpu_test
 
 import (
+	"errors"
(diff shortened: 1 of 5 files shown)


jaypipes (approved):
Nice @fromanirh :)

PR #236, opened 2021-04-11
context: Fix directory leakage for snapshots contexts

marcel-apf:
Fix an error preventing the auto-unpacked snapshot files to be deleted.
Add a unit-test to expose the issue.

Signed-off-by: Marcel Apfelbaum <marcel@redhat.com>

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -14,12 +14,13 @@ import (
 // Concrete merged set of configuration switches that act as an execution
 // context when calling internal discovery methods
 type Context struct {
-	Chroot            string
-	EnableTools       bool
-	SnapshotPath      string
-	SnapshotRoot      string
-	SnapshotExclusive bool
-	alert             option.Alerter
+	Chroot               string
+	EnableTools          bool
+	SnapshotPath         string
+	SnapshotRoot         string
+	SnapshotExclusive    bool
+	snapshotUnpackedPath string
+	alert                option.Alerter
 }
 
 // New returns a Context struct pointer that has had various options set on it
@@ -92,6 +93,9 @@ func (ctx *Context) Setup() error {
 	root := ctx.SnapshotRoot
 	if root == "" {
 		root, err = snapshot.Unpack(ctx.SnapshotPath)
+		if err == nil {
+			ctx.snapshotUnpackedPath = root
+		}
 	} else {
 		var flags uint
 		if ctx.SnapshotExclusive {
@@ -111,12 +115,12 @@ func (ctx *Context) Setup() error {
 // You should always call `Teardown` if you called `Setup` to free any resources
 // acquired by `Setup`. Check `Do` for more automated management.
 func (ctx *Context) Teardown() error {
-	if ctx.SnapshotRoot != "" {
+	if ctx.snapshotUnpackedPath == "" {
 		// if the client code provided the unpack directory,
 		// then it is also in charge of the cleanup.
 		return nil
 	}
-	return snapshot.Cleanup(ctx.SnapshotRoot)
+	return snapshot.Cleanup(ctx.snapshotUnpackedPath)
 }
 
 func (ctx *Context) Warn(msg string, args ...interface{}) {
--- a/pkg/context/context_test.go
+++ b/pkg/context/context_test.go
@@ -1,0 +1,42 @@
+//
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
+package context_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/jaypipes/ghw/pkg/context"
+	"github.com/jaypipes/ghw/pkg/option"
+)
+
+const (
+	testDataSnapshot = "../snapshot/testdata.tar.gz"
+)
+
+// nolint: gocyclo
+func TestSnapshotContext(t *testing.T) {
+	ctx := context.New(option.WithSnapshot(option.SnapshotOptions{
+		Path: testDataSnapshot,
+	}))
+
+	var uncompressedDir string
+	err := ctx.Do(func() error {
+		uncompressedDir = ctx.Chroot
+		return nil
+	})
+
+	if uncompressedDir == "" {
+		t.Fatalf("Expected the uncompressed dir path to not be empty")
+	}
+	if err != nil {
+		t.Fatalf("Expected nil err, but got %v", err)
+	}
+	if _, err = os.Stat(uncompressedDir); !os.IsNotExist(err) {
+		t.Fatalf("Expected the uncompressed dir to be deleted: %s", uncompressedDir)
+	}
+}


fromanirh:
thanks for the PR. We indeed leak the unpacked snapshot directory. it seems we do only on `Unpack`. There is also an interesting interplay with nested context. Given how `ghw` uses snapshots, this is only in the test path and it seems to happen only when `ghw` owns the snapshot unpack directory.

fromanirh:
This fix will expose a bug in a couple tests, we'll need to fix them first - I'll post a PR soon(ish).

pkg/context/context.go
```diff
@@ -116,7 +116,7 @@ func (ctx *Context) Teardown() error {
 		// then it is also in charge of the cleanup.
 		return nil
 	}
-	return snapshot.Cleanup(ctx.SnapshotRoot)
+	return snapshot.Cleanup(ctx.Chroot)
```
fromanirh, line 119: In general I don't think we should clean up `Chroot`. While this should work in the happy path, we leave too much room for error: for example we need to ensure that we never try to clean up user-supplied directory with user data.
Now: the original code was ALSO wrong, and for extra fun the doc is also a bit misleading. I'll fix the latter with a separate PR.

pkg/context/context_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+// nolint: gocyclo
+func TestSnapshotContext(t *testing.T) {
+	ctx := context.New(option.WithSnapshot(option.SnapshotOptions{
+		Path: testDataSnapshot,
+	}))
+
+	var uncompressedDir string
+	err := ctx.Do(func() error {
+		uncompressedDir = ctx.Chroot
```
fromanirh, line 29: the test itself looks good though.

pkg/context/context.go
fromanirh: https://github.com/jaypipes/ghw/pull/239

marcel-apf: agreed, looks like `Chroot` choice may be too risky. I was going to add another flag to ctx (cleanupNeeded) and use it for cleanup, but it didn't look elegant enough.

fromanirh: I'd store the internally-managed directory explicitely, like in...
```go
type Context struct {
	Chroot            string
	EnableTools       bool
	SnapshotPath      string
	SnapshotRoot      string // set by user
	SnapshotExclusive bool
	alert             option.Alerter
	parent            *Context
	snapshotRoot      string // managed internally
}
//...
func (ctx *Context) Setup() error {
	if ctx.SnapshotPath == "" {
		// nothing to do!
		return nil
	}

	var err error
	root := ctx.SnapshotRoot
	if root == "" {
		root, err = snapshot.Unpack(ctx.SnapshotPath)
		ctx.snapshotRoot = root
	} else {
		var flags uint
		if ctx.SnapshotExclusive {
			flags |= snapshot.OwnTargetDirectory
		}
		_, err = snapshot.UnpackInto(ctx.SnapshotPath, root, flags)
	}
	if err != nil {
		return err
	}

	ctx.Chroot = root
	return nil
}

func (ctx *Context) Teardown() error {
	if ctx.snapshotRoot == "" {
		// if the client code provided the unpack directory,
		// then it is also in charge of the cleanup.
		return nil
	}
	return snapshot.Cleanup(ctx.snapshotRoot)
}
```

marcel-apf: yes,  maybe indeed it is worth adding a new internal field just to be safe. I'll post the changes soon, thanks.

fromanirh: Thanks! no hurry. And with my https://github.com/jaypipes/ghw/pull/239 merged, is clear that ghw wants to clean only the directory it owns - hence when `GHW_SNAPSHOT_ROOT` is *not* given. This should help a bit with this PR as well.

marcel-apf: I re-pushed following your comments, and yes, now the code is cleaner, thanks for the review.

pkg/context/context_test.go
marcel-apf: great, thanks

fromanirh:
The test failures are expected (https://github.com/jaypipes/ghw/pull/236#issuecomment-817319852) and fixed by https://github.com/jaypipes/ghw/pull/238/commits/b7b2bc6409e3f0f7d47464e2239a890ef76719e8
We can probably extract a more self contained fix and add to this PR. I can rebase my 238 later on.

pkg/context/context.go
```diff
@@ -111,12 +115,10 @@ func (ctx *Context) Setup() error {
 // You should always call `Teardown` if you called `Setup` to free any resources
 // acquired by `Setup`. Check `Do` for more automated management.
 func (ctx *Context) Teardown() error {
-	if ctx.SnapshotRoot != "" {
-		// if the client code provided the unpack directory,
-		// then it is also in charge of the cleanup.
+	if len(ctx.snapshotUnpackedPath) == 0 {
```
fromanirh, line 118: mostly curious, why just not
```go
if ctx.snapshotUnpackedPath == "" {
```

fromanirh:
https://github.com/jaypipes/ghw/pull/240 is the minimal fix. It addresses only the lifecycle of the unpacked snapshot, same thing we are doing here. 

https://github.com/jaypipes/ghw/pull/238 is a bigger change to really enable more efficient (and IMHO, logical/consistent) snapshot reuse, which we may want to postpone.

pkg/context/context.go
marcel-apf: because of a personal preference, I don't like to use string literals. If golang would have had a strings.Empty constant and specify it is the default string value, it would be fine, otherwise it has a "smell" (using literlas in code).
The joke is on me since the integer 0 is also a literal.

You can also have a look on the 'tabs vs spaces' like (https://www.youtube.com/watch?v=SsoOG6ZeyUI) "check empty strings" google  chat:
   https://groups.google.com/g/golang-nuts/c/7Ks1iq2s7FA?pli=1
here is an interesting quote from Russ Cox extracted from the above discussion:
 `
The one that makes the code clear.
If I'm about to look at element x I typically write
len(s) > x, even for x == 0, but if I care about
"is it this specific string" I tend to write s == "".
It's reasonable to assume that a mature compiler will compile
len(s) == 0 and s == "" into the same, efficient code.
Right now 6g etc do compile s == "" into a function call
while len(s) == 0 is not, but that's been on my to-do list to fix.
`

fromanirh: thanks, makes sense. I don't have a strong opinion here so I don't mind.

fromanirh:
I think this is a good fix. Thanks for the updates!

pkg/context/context.go
```diff
@@ -111,12 +115,10 @@ func (ctx *Context) Setup() error {
 // You should always call `Teardown` if you called `Setup` to free any resources
 // acquired by `Setup`. Check `Do` for more automated management.
 func (ctx *Context) Teardown() error {
-	if ctx.SnapshotRoot != "" {
-		// if the client code provided the unpack directory,
-		// then it is also in charge of the cleanup.
```
jaypipes, line 116: Hi @marcel-apf, can you please put the above code comment back into the file? It helps to explain why we're doing this...

pkg/context/context.go
jaypipes: @marcel-apf my rule of thumb is to generally use the style of code already present in the codebase :) If you wouldn't mind using the `== ""` style, I'd appreciate that.

jaypipes (changes requested):

pkg/context/context.go
marcel-apf: done

pkg/context/context.go
marcel-apf: done

jaypipes (approved):
thanks @marcel-apf! :)

PR #202, opened 2020-12-14
pkg: add support to consume snapshots

fromanirh:
Needed for https://github.com/jaypipes/ghw/issues/66 (see https://github.com/jaypipes/ghw/issues/66#issuecomment-738618640 and following)

Add support to allow `ghw` read from snapshots (created with `ghw-snapshot`) using
environment variables or programmatically.
This is meant to be used mostly in the testsuite, but can be useful for
users in general.

Added environment variables (and counterpart in golang pkgs):
- `GHW_SNAPSHOT_PATH` let users specify a snapshot
   that `ghw` will automatically consume.
- `GHW_SNAPSHOT_ROOT` let users specify the directory
   on which the snapshot should be unpacked.
- `GHW_SNAPSHOT_PRESERVE` is relevant iff `GHW_SNAPSHOT_ROOT` is given.
   Tells `ghw` that the directory is meant only to contain the given snapshot,
   thus `ghw` will *not* attempt to unpack it
   (and will go ahead silently!) unless the directory is empty.
- `GHW_SNAPSHOT_KEEP` if set, `ghw` will *not*
   clean up the unpacked snapshot once done, leaving it into the system.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #230, opened 2021-02-24
expose SRIOV information

ffromani:
Expose informations about SRIOV devices. After some design iterations, we fixed these goals:
- model SRIOV devices like GPUs, so with a top-level package.
- to shape the API, we start checking what the [k8s SRIOV operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) needs
- to further refine later, we will check what other projects like the [k8s node feature discovery](https://github.com/kubernetes-sigs/node-feature-discovery) consume
- the goal is to enable such projects to switch to ghw if/when they want.

There are few more noteworthy items in this PR:
- due to lack of resources, initial support is for linux only. Nothing linux-specific should have sneaked in the API. Should the current API design make unnecessarily hard to add support on other platforms (e.g. windows) this should be treated as bug
- the code prefers Physical Functions for discovery, but we acknowledge that forcing consumers to traverse the physical functions to learn about virtual functions is awkward, so the API provides shallow references to the Virtual Functions for the sake of practicality.

Fixes: https://github.com/jaypipes/ghw/issues/92

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #238, opened 2021-04-11
context: introduce dependent context

fromanirh:
The PR https://github.com/jaypipes/ghw/pull/236 fixes a snapshot
directory leakage issue, but also highlight a context
lifecycle/ownership issue.

In general in ghw each package creates its own context, including
the cases on which a package is consumed by another, for example
`topology` and `pci`. However, in this case, it would make more sense
to reuse the context from the parent package, because the two packages
are tightly coupled.

Before the introduction of the the transparent snapshot support
(https://github.com/jaypipes/ghw/pull/202), the above mechanism worked,
because each context, each time, was consuming over and over again the
same parameters (e.g. environment variables); besides some resource
waste, this had no negative effect.

When introducing snapshots in the picture, repeatedly unpacking the
same snapshot to consume the same data is much more wasteful.
So it makes sense to introduce explicitly the concept of dependent
context.

We add a functio to create a context subordinate to another.
The subordinate context will effectively borrow all the resources from
the parent one. The parent is in charge to perform any setup/teardown
(e.g. for snapshot, to learn which chroot should be used, if at all)
and so forth.
Making this dependency explicit allow the client code to manage
correctly the lifecycle of the topmost context, hence of all the
resources.

### DIFF
--- a/pkg/baseboard/baseboard.go
+++ b/pkg/baseboard/baseboard.go
@@ -50,7 +50,14 @@ func (i *Info) String() string {
 // New returns a pointer to an Info struct containing information about the
 // host's baseboard
 func New(opts ...*option.Option) (*Info, error) {
-	ctx := context.New(opts...)
+	return newWithContext(context.New(opts...))
+}
+
+func NewWithContext(ctx *context.Context) (*Info, error) {
+	return newWithContext(context.FromParent(ctx))
+}
+
+func newWithContext(ctx *context.Context) (*Info, error) {
 	info := &Info{ctx: ctx}
 	if err := ctx.Do(info.load); err != nil {
 		return nil, err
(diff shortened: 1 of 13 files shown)


fromanirh:
@jaypipes a bunch of cleanups found while helping with https://github.com/jaypipes/ghw/pull/236
I know I need to add a lot of docstrings in https://github.com/jaypipes/ghw/pull/238/commits/b4b6650a5be089cfe357ee8511c5765d1a935fa1 :) But please review so we can discuss the concepts exposed in this PR.

fromanirh:
the SRIOV support (https://github.com/jaypipes/ghw/pull/230) is the main driver for this PR, so I think I'll just rebase 230 on top of this one.

fromanirh:
done, let's continue on https://github.com/jaypipes/ghw/pull/230

PR #237, opened 2021-04-11
test: snapshot: don't leak unpack directory

fromanirh:
Errors in the tests caused leaks of unpacked test directory.
Make sure to clean up after ourselves.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #239, opened 2021-04-11
doc: don't claim we clean up user-supplied dirs

fromanirh:
Previously the documentation claimed the snapshot support code
could clean up properly when `GHW_SNAPSHOT_ROOT` is given.
This is too risky to be done safely, and having shared ownership
(client code owns the directory, ghw owns _some_ of its content)
is a bad idea anyway.

Thus, better not to attempt half-baked work: ghw will leave the
GHW_SNAPSHOT_ROOT directory alone completely.

Please note this also reflects what the code actually does
and the intention expressed in the comment in `pkg/context/context.go`


jaypipes (approved):
++ we only do this on tests now with your PR #237

PR #235, opened 2021-03-30
gpu: test to handle missing drm numa node (issue#234)

fromanirh:
WIP - details to follow

PR #243, opened 2021-05-03
snapshot: collect data related to GPU devices

fromanirh:
make sure the snapshot code collects data related to GPU devices.
Turns out not much was missing, most notably the `/sys/class/drm` link;
nevertheless, let's make sure to collect this data, so now also the
gpu package can consume snapshots.

We choose to refactor the code from net devices collection more
for clarity -what we collect and what we filter out is a bit more
explicit now- than to avoid duplication.

The PCI collection code is a bit more elaborate, because it needs to
deal with PCI roots and subtrees (subforests?), so it was intentionally
left out for the time being.

### DIFF
--- a/pkg/snapshot/clonetree.go
+++ b/pkg/snapshot/clonetree.go
@@ -52,6 +52,7 @@ func ExpectedCloneContent() []string {
 	fileSpecs := ExpectedCloneStaticContent()
 	fileSpecs = append(fileSpecs, ExpectedCloneNetContent()...)
 	fileSpecs = append(fileSpecs, ExpectedClonePCIContent()...)
+	fileSpecs = append(fileSpecs, ExpectedCloneGPUContent()...)
 	return fileSpecs
 }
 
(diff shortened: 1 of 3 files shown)


fromanirh:
split from https://github.com/jaypipes/ghw/pull/235

pkg/snapshot/clonetree.go
```diff
@@ -202,3 +203,55 @@ func copyLink(path, targetPath string) error {
 
 	return nil
 }
+
+// cloneContentByClass copies all the content related to a given device class (devClass), possibly filtering out devices whose name
+// does NOT pass a filter (filterName). Each entry in `/sys/class/$CLASS` is actually a symbolic link. We can filter out entries depending
+// on the link target. Each filter is a simple function which takes the entry name or the link target and must return true if the
+// entry should be collected, false otherwise. last but not list, explicitely collect a list of attributes for each entry, given
+// as list of glob patterns as `subEntries`. Return the final list of glob patterns to be collected.
+func cloneContentByClass(devClass string, subEntries []string, filterName func(string) bool, filterLink func(string) bool) []string {
```
jaypipes, line 212: suggestion: make a typedef for the filter function:

```go
type FilterFunc func(string) bool
```

and then use that typedef in this function signature:

```go
func cloneContentByClass(
    devClass string,
    subEntries []string,
    filterName FilterFunc,
    filterLink FilterFunc,
) []string {
...
}
```

jaypipes (changes requested):
Small nits inline, otherwise looks great, thank you @fromanirh!

pkg/snapshot/clonetree.go
```diff
@@ -202,3 +203,55 @@ func copyLink(path, targetPath string) error {
 
 	return nil
 }
+
+// cloneContentByClass copies all the content related to a given device class (devClass), possibly filtering out devices whose name
```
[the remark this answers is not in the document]
fromanirh: ack, done

pkg/snapshot/clonetree.go
fromanirh: ack, done

jaypipes (approved):
w00t :) thanks @fromanirh!

PR #244, opened 2021-05-03
gpu: pass through the context into subsystems

fromanirh:
The GPU package uses the pci and the topology packages internally.
Make sure to pass to them the same context the package is using,
in order to ensure more consistent behaviour, enable snapshot
consumption - and to waste a little less work in general.


fromanirh:
split from https://github.com/jaypipes/ghw/pull/235

PR #247, opened 2021-05-13
Dependent contexts

fromanirh:
In this PR we revisit how we handle context for dependent packages; a good example is the `pci` package, which wants to use the `topology` package internally. But there are more, like `gpu` (uses `pci` and `topology`), and in the future `sriov` wants as well to use `pci` (see https://github.com/jaypipes/ghw/pull/230)

While I'm happy with this PR and I believe the gains outweights the risk, let's discuss if this direction is a good one or if we want to solve the same problem in a different way.

### DIFF
--- a/pkg/baseboard/baseboard.go
+++ b/pkg/baseboard/baseboard.go
@@ -57,7 +57,14 @@ func (i *Info) String() string {
 // New returns a pointer to an Info struct containing information about the
 // host's baseboard
 func New(opts ...*option.Option) (*Info, error) {
-	ctx := context.New(opts...)
+	return NewWithContext(context.New(opts...))
+}
+
+// New returns a pointer to an Info struct containing information about the
+// host's baseboard, reusing a given context.
+// Use this function when you want to consume this package from another,
+// ensuring the two see a coherent set of resources.
+func NewWithContext(ctx *context.Context) (*Info, error) {
 	info := &Info{ctx: ctx}
 	if err := ctx.Do(info.load); err != nil {
 		return nil, err
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -7,6 +7,8 @@
 package context
 
 import (
+	"fmt"
+
 	"github.com/jaypipes/ghw/pkg/option"
 	"github.com/jaypipes/ghw/pkg/snapshot"
 )
@@ -23,6 +25,7 @@ type Context struct {
 	snapshotUnpackedPath string
 	alert                option.Alerter
 	refCount             int
+	err                  error
 }
 
 // New returns a Context struct pointer that has had various options set on it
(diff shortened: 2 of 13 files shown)


fromanirh:
test failures are unexpected, let me fix them before we move further with the review.

jaypipes (changes requested):
@fromanirh I'm a little confused why you want to embed a parent context inside Context... see my comment inline :)

pkg/baseboard/baseboard.go
```diff
@@ -57,7 +57,18 @@ func (i *Info) String() string {
...
-	ctx := context.New(opts...)
+	return newWithContext(context.New(opts...))
+}
+
+// New returns a pointer to an Info struct containing information about the
+// host's baseboard, reusing a given context.
+// Use this function when you want to consume this package from another,
+// ensuring the two see a coherent set of resources.
+func NewWithContext(ctx *context.Context) (*Info, error) {
+	return newWithContext(context.FromParent(ctx))
```
jaypipes, line 68: I don't see a need for the `context.FromParent` function or to embed one context in another. Why not just set the context you receive directly into the Info struct? In other words, just do this (shown for the CPU package...)?

```go
// New returns a pointer to an Info struct that contains information about the
// CPUs on the host system
func New(opts ...*option.Option) (*Info, error) {
    return NewWithContext(context.New(opts...))
}

// NewWithContext returns a pointer to an Info struct containing information
// about the CPUs on the host system given an existing Context 
func NewWithContext(ctx *context.Context) (*Info, error) {
    info := &amp;Info{ctx: ctx}
    if err := ctx.Do(info.load); err != nil {
        return nil, err
    }
    return info, nil
}
```

jaypipes (changes requested):

pkg/baseboard/baseboard.go
fromanirh: This is a very good point. Thinking about it, we actually don't. What we actually need, is a mechanism for shared contexts to know when Setup() and Teardown() must happen (= the outermost package must do that, the dependent contexts must not).
So we need a form of reference counting. Let me update the code.

fromanirh:
sorry for the churn, uploaded a botched version by mistake, should be fixed now.

pkg/context/context.go
```diff
@@ -53,10 +57,17 @@ func New(opts ...*option.Option) *Context {
 		ctx.PathOverrides = merged.PathOverrides
 	}
 
+	// New is not allowed to return error - it would break the established API.
+	// so the only way out is to actually do the checks here and record the error,
+	// and return it later, at the earliest possible occasion, in Setup()
+	if ctx.SnapshotPath != "" && ctx.Chroot != option.DefaultChroot {
+		// The env/client code supplied a value, but we are will overwrite it when unpacking shapshots!
+		ctx.err = fmt.Errorf("Conflicting options: chroot %q and snapshot path %q", ctx.Chroot, ctx.SnapshotPath)
```
jaypipes, line 65: @fromanirh instead of the error tracking and refcounting, why don't we simply `panic()` here? I think `panic()` is entirely logical considering this situation would occur only when the user has explicitly made a mistake in their argument-passing.

fromanirh: just a note: refcounting is meant to solve a different facet of the problem (handling nested context in such a way we perform the cleanup in the expected order)
In general I prefer to avoid to panic() inside packages, this should always be an application layer decision - not a library layer decision, so I'd really try hard to find another approach if you don't like the admittedly awkward way I'm using here.
Let me look for other options

PR #248, opened 2021-05-14
linuxpaths: allow to override mount points

fromanirh:
Inside containers, when granting access to the host file systems, is
common practice to bind-mount specific host subtrees on nonstandard
locations, like `/sys` on `/host-sys` or `/proc` on `/host-proc`.

To better cover this use case, `ghw` allows to *programmatically* override
the initial component of filesystems subtrees, allowing to access `sysfs`
(or `procfs` or...) mounted on these non-standard locations.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -32,9 +32,12 @@ var (
 	// match the existing environ variable to minimize surprises
 	WithDisableWarnings = option.WithNullAlerter
 	WithDisableTools    = option.WithDisableTools
+	WithPathOverrides   = option.WithPathOverrides
 )
 
 type SnapshotOptions = option.SnapshotOptions
+
+type PathOverrides = option.PathOverrides
 
 type CPUInfo = cpu.Info
 
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -19,6 +19,7 @@ type Context struct {
 	SnapshotPath         string
 	SnapshotRoot         string
 	SnapshotExclusive    bool
+	PathOverrides        option.PathOverrides
 	snapshotUnpackedPath string
 	alert                option.Alerter
 }
@@ -46,6 +47,10 @@ func New(opts ...*option.Option) *Context {
 
 	if merged.EnableTools != nil {
 		ctx.EnableTools = *merged.EnableTools
+	}
+
+	if merged.PathOverrides != nil {
+		ctx.PathOverrides = merged.PathOverrides
 	}
 
 	return ctx
--- a/pkg/linuxpath/path_linux.go
+++ b/pkg/linuxpath/path_linux.go
@@ -12,6 +12,49 @@ import (
 	"github.com/jaypipes/ghw/pkg/context"
 )
 
+// PathRoots holds the roots of all the filesystem subtrees
+// ghw wants to access.
+type PathRoots struct {
+	Etc  string
+	Proc string
+	Run  string
+	Sys  string
+	Var  string
+}
+
+// DefaultPathRoots return the canonical default value for PathRoots
+func DefaultPathRoots() PathRoots {
+	return PathRoots{
+		Etc:  "/etc",
+		Proc: "/proc",
+		Run:  "/run",
+		Sys:  "/sys",
+		Var:  "/var",
+	}
+}
+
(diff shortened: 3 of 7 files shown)


pkg/context/context.go
```diff
@@ -19,6 +19,11 @@ type Context struct {
 	SnapshotPath         string
 	SnapshotRoot         string
 	SnapshotExclusive    bool
+	PathEtc              string
+	PathProc             string
+	PathRun              string
+	PathSys              string
+	PathVar              string
```
jaypipes, line 26: These are all Linux-specific and should be in `pkg/linuxpath` package.

For the generic context package, you can do something like this in `pkg/option/option.go`:

```go
// PathOverrides is a map, keyed by the string name of a mount path, of override paths
type PathOverrides map[string]string

// WithPathOverrides supplies path-specific overrides for the context
func WithPathOverrides(overrides PathOverrides) *Option {
	return &Option{
        		PathOverrides: overrides,
       	}
}
```

jaypipes (commented):
suggestion inline about keeping linux-specific paths in `pkg/linuxpath`

pkg/context/context.go
fromanirh: You are right. The only reason why I put them here is to (try to) avoid the circular dependency issue. I'll try out the approach you suggested, which is cleaner indeed.

alias.go
```diff
@@ -32,10 +32,13 @@ var (
 	// match the existing environ variable to minimize surprises
 	WithDisableWarnings = option.WithNullAlerter
 	WithDisableTools    = option.WithDisableTools
+	WithPathOverrides   = option.WithPathOverrides
 )
 
 type SnapshotOptions = option.SnapshotOptions
 
+type PathOverridesOptions = option.PathOverrides
```
jaypipes, line 40: Let's call the alias `PathOverrides` instead of `PathOverridesOptions` in order to match the existing alias scheme.

pkg/linuxpath/path_linux.go
```diff
@@ -12,6 +12,49 @@ import (
...
+		roots.Run = pathRun
+	}
+	if pathSys, ok := ctx.PathOverrides["/sys"]; ok {
+		roots.Sys = pathSys
+	}
+	if pathVar, ok := ctx.PathOverrides["/var"]; ok {
+		roots.Var = pathVar
+	}
+	return roots
+}
```
jaypipes, line 56: :+1:

pkg/linuxpath/path_linux.go
```diff
@@ -31,20 +74,21 @@ type Paths struct {
...
+		SysKernelMMHugepages:   filepath.Join(ctx.Chroot, roots.Sys, "kernel", "mm", "hugepages"),
+		EtcMtab:                filepath.Join(ctx.Chroot, roots.Etc, "mtab"),
+		SysBlock:               filepath.Join(ctx.Chroot, roots.Sys, "block"),
+		SysDevicesSystemNode:   filepath.Join(ctx.Chroot, roots.Sys, "devices", "system", "node"),
+		SysDevicesSystemMemory: filepath.Join(ctx.Chroot, roots.Sys, "devices", "system", "memory"),
+		SysBusPciDevices:       filepath.Join(ctx.Chroot, roots.Sys, "bus", "pci", "devices"),
+		SysClassDRM:            filepath.Join(ctx.Chroot, roots.Sys, "class", "drm"),
+		SysClassDMI:            filepath.Join(ctx.Chroot, roots.Sys, "class", "dmi"),
+		SysClassNet:            filepath.Join(ctx.Chroot, roots.Sys, "class", "net"),
+		RunUdevData:            filepath.Join(ctx.Chroot, roots.Run, "udev", "data"),
```
jaypipes, line 91: nice. :)

jaypipes (changes requested):
one teensy-tiny rename requested, otherwise awesome :)

alias.go
fromanirh: sure thing!

pkg/linuxpath/path_linux.go
fromanirh: :)

jaypipes (approved):
Thank you @fromanirh!

PR #251, opened 2021-05-16
WIP: block: read procfs, not /etc/mtab

fromanirh:
docs TBD


fromanirh:
reason for this change: https://github.com/systemd/systemd/blob/v227/NEWS#L9-L14

pkg/block/block_linux.go
```diff
@@ -410,15 +410,15 @@ func partitionInfo(paths *linuxpath.Paths, part string) (string, string, bool) {
 	return "", "", true
 }
 
-type mtabEntry struct {
+type mountInfoEntry struct {
```
jaypipes, line 413: we could probably just call this `mountEntry`?

fromanirh: sure thing

jaypipes (approved):
thank you sir! :)

Issue #220, opened 2021-01-25
PCI: Migrate from BSF to BDF convention


pearsonk:
Slot designation is a bit of a misnomer when referring to the addressing field, which is conventionally called Device in Domain:Bus:Device.Function notations.
Further, as it seems relevant, actual slot information can be pulled from DMI (type 9) with a lookup of the Designation for a matching Bus Address for the first Function of the device. Absence of a DMI entry would signify an internal PCI device not connected to an expansion slot.

fromanirh:
fixed by https://github.com/jaypipes/ghw/pull/252

PR #252, opened 2021-05-20
rename PCIAddress.Slot to PCIAddress.Device

jaypipes:
As noted by @pearsonk in #220, the PCIAddress.Slot field was a bit of a
misnomer, since the more common representation of a PCIAddress is
Bus:Device.Function, not Bus:Slot.Function. Actual PCI slot information
may be retrieved from DMI, but what we were referring to was not the
actual PCI slot information but rather the PCI address device component.

Fixes Issue #220

### DIFF
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -40,7 +40,7 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
 	}
 	return filepath.Join(
 		paths.SysBusPciDevices,
-		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Slot+"."+pciAddr.Function,
+		pciAddr.String(),
 		"modalias",
 	)
 }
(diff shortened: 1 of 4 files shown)


pkg/pci/pci_linux.go
```diff
@@ -40,7 +40,7 @@ func getDeviceModaliasPath(ctx *context.Context, address string) string {
 	}
 	return filepath.Join(
 		paths.SysBusPciDevices,
-		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Slot+"."+pciAddr.Function,
+		pciAddr.Domain+":"+pciAddr.Bus+":"+pciAddr.Device+"."+pciAddr.Function,
```
fromanirh, line 43: We should just use `pciAddr.String()` here - perhaps in a later patch

fromanirh (commented):
minor suggestion inside, otherwise lgtm

pkg/pci/pci_linux.go
jaypipes: ++ fixed.

fromanirh:
Looks great to me :)

PR #254, opened 2021-05-31
travis: add more test lanes

fromanirh:
Let's test more recent golang versions.


fromanirh:
We plan to move to GH actions; so let's re-evaluate after we moved.

PR #258, opened 2021-06-11
block: honour EnableTools flag

fromanirh:
turns out we call external tools in the block subpackage, hence
we need to handle the EnableTools flag we added in commit
591a1bababf9ac1446a13e75ab3cae2df40b9619

### DIFF
--- a/pkg/block/block_darwin.go
+++ b/pkg/block/block_darwin.go
@@ -207,6 +207,10 @@ func storageControllerFromPlist(infoPlist *diskUtilInfoPlist) StorageController 
 }
 
 func (info *Info) load() error {
+	if !info.ctx.EnableTools {
+		return fmt.Errorf("EnableTools=false on darwin disables block support entirely.")
+	}
+
 	listPlist, err := getDiskUtilListPlist()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
--- a/pkg/block/block_linux.go
+++ b/pkg/block/block_linux.go
@@ -210,6 +210,10 @@ func diskPartitions(ctx *context.Context, paths *linuxpath.Paths, disk string) [
 }
 
 func diskPartUUID(ctx *context.Context, part string) string {
+	if !ctx.EnableTools {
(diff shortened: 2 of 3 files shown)


fromanirh:
let's use this simple WIP PR to try out the GH actions scoring for PRs :)

pkg/block/block_darwin.go
```diff
@@ -207,6 +207,11 @@ func storageControllerFromPlist(infoPlist *diskUtilInfoPlist) StorageController
 }
 
 func (info *Info) load() error {
+	if !info.ctx.EnableTools {
+		// we need to be verbose here
+		info.ctx.Warn("Skip EnableTools=false on darwin: would disable block support entirely.")
```
jaypipes, line 212: This isn't actually skipping anything, though, right? I mean, we're not returning after this warning...

pkg/block/block_linux.go
```diff
@@ -210,6 +210,13 @@ func diskPartitions(ctx *context.Context, paths *linuxpath.Paths, disk string) [
 }
 
 func diskPartUUID(ctx *context.Context, part string) string {
+	if !ctx.EnableTools {
+		// intentionally not using ctx.Warn because
+		// 1. we want to avoid excessive message flow
+		// 2. this is an explicit setting, which client code
+		//    must be prepared to handle.
+		return ""
+	}
```
jaypipes, line 219: I guess I'm not entirely sure why we're retunrning `""` here and not outputting at least a warning message, since this is changing the behaviour of `ghw` without any indication to the user. We warn below on original line 224 and 234 if we encounter an error using the tool `blkid` so I'm unsure why this isn't similar?

pkg/block/block_darwin.go
fromanirh: yep, got lost in the rebase :\
my fault, will fix

pkg/block/block_linux.go
fromanirh: good point. This is a plain old mistake on my side. Will revisit and re-upload.

jaypipes (approved):
:+1: thanks @fromanirh!

PR #259, opened 2021-06-11
gh actions: add more lanes

fromanirh:
With this change, we reach parity with the travis lanes
we planned to add. We replace older go version with more modern ones.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #260, opened 2021-06-14
snapshot: mark OS-specific code as so

fromanirh:
In the snapshot package we have code to clone relevant filesystem
information. This is by its nature OS-specific, but because a oversight
we previously didn't marked as so.

This became evident while trying to enable darwin and windows specific
CI lanes (see https://github.com/jaypipes/ghw/pull/259).

With this patch we now correctly declare the clonetree code as OS
specific. No intended changes in behaviour on linux.

As final note, this change also highlight we will need snapshot-specific
code for darwin/windows (or a similar machinery), which we postpone
for lack of capacity and access to relevant platforms.


jaypipes (approved):
Nice :)

PR #261, opened 2021-06-15
clonetree: stub: fix nonlinux builds

fromanirh:
Fix the builds removing leftover imprts, and mark actually specific linux test as so.

The existing clonetree tests are linux specific because  they weere developed on linux.
While no intentional restriction on linux was put in, no real thought to other OSes was put in either.

So we just make the expectations explicit and we mark these tests as linux specific until we have the chance to truly generalize them.


fromanirh:
now the darwin lane is green. We do little in this lane, but this is the best we can do atm. :\

fromanirh:
Same goes for the windows lane. Most tests disabled. I guess is the best we can do with so limited platform availability

fromanirh:
Once this is merged, we can also merge https://github.com/jaypipes/ghw/pull/262

jaypipes (approved):
Thanks Francesco! :)

PR #265, opened 2021-06-26
pci: check against corrupted modalias data

fromanirh:
Add a sanity check against bad modalias data (corrupted/truncated).
Thsi should never happen in practice, but it's a trivial check,
prevents a nasty segfault and looks like the sriov device plugin
checks for this case (https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin/blob/v3.3.2/cmd/sriovdp/manager_test.go#L322)
so let's add it.

### DIFF
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -20,6 +20,11 @@ import (
 	"github.com/jaypipes/ghw/pkg/util"
 )
 
+const (
+	// found running `wc` against real linux systems
+	modAliasExpectedLength = 54
+)
+
 func (i *Info) load() error {
 	db, err := pcidb.New(pcidb.WithChroot(i.ctx.Chroot))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/pci/pci_linux.go
```diff
@@ -112,6 +117,10 @@ func parseModaliasFile(fp string) *deviceModaliasInfo {
 }
 
 func parseModaliasData(data string) *deviceModaliasInfo {
+	// extra sanity check to avoid segfaults
+	if len(data) < modAliasExpectedLength {
```
jaypipes, line 121: what if the data is greater than 54 chars? :)

fromanirh: We should happily ignore the extra data, but this calls for both a comment and a test! Will add them!

jaypipes (approved):
thanks @fromanirh! :)

PR #268, opened 2021-08-26
memory: topology: expose per-numa memory

fromanirh:
This patch wants to add memory informations on topology.Node,
so we can have per-NUMA-zone memory information.

WIP: tests are broken (and more needs to come), doc is missing,
API is provisional. The basic concepts are there, though.


fromanirh:
@jaypipes I think this PR is ready for another round of review. I removed all the hugepage reporting (it was at very least too premature, sorry for the noise) and fixed the pending items I had.

fromanirh:
@jaypipes - not urgent at all, but I think this PR is ready for review, so just keeping it on our radar

pkg/memory/memory_linux.go
```diff
@@ -52,10 +53,56 @@ func (i *Info) load() error {
...
+	}
+
+	totUsable, err := memoryTotalUsableBytesFromPath(filepath.Join(path, "meminfo"))
+	if err != nil {
+		return nil, err
+	}
+
+	supportedHP, err := memorySupportedPageSizes(filepath.Join(path, "hugepages"))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "err=%v\n", err)
```
jaypipes, line 84: Is there a reason we are only printing to stderr on this particular error and not the other errors above? Can we call `ctx.Warn` instead for all four subcalls' error returns?

fromanirh: this is because it's a debug leftover :) you raised a good point though, I'll check if/how I can add a `ctx.Warn` here.

fromanirh: OK, in this case is not worth to add a `ctx.Warn` IMHO, so I'm just dropping the `Fprintf` call. However, I found some cleanups to be done and I'll add to the PR.

jaypipes (approved):
Great work, @fromanirh :) :+1:

fromanirh:
Thanks!

PR #277, opened 2021-09-20
Rename withAlterter to withAlerter

GreyXor:
There was an misspelling in alias.go.
I propose to :

- rename `alterter` to `alerter`.
- add the old `withAlterter` (for compatibility reasons) but put them deprecated.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -27,8 +27,8 @@ type WithOption = option.Option
 var (
 	WithChroot       = option.WithChroot
 	WithSnapshot     = option.WithSnapshot
-	WithAlerter     = option.WithAlerter
-	WithNullAlerter = option.WithNullAlerter
+	WithAlerter      = option.WithAlerter
+	WithNullAlerter  = option.WithNullAlerter
 	// match the existing environ variable to minimize surprises
 	WithDisableWarnings = option.WithNullAlerter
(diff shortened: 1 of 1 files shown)


alias.go
```diff
@@ -27,12 +27,16 @@ type WithOption = option.Option
...
+	WithAlerter      = option.WithAlerter
+	WithNullAlerter  = option.WithNullAlerter
 	// match the existing environ variable to minimize surprises
 	WithDisableWarnings = option.WithNullAlerter
 	WithDisableTools    = option.WithDisableTools
 	WithPathOverrides   = option.WithPathOverrides
+	// Deprecated: WithAlterter is misspelled, Instead, use WithAlerter.
+	WithAlterter      = option.WithAlerter
+	// Deprecated: WithNullAlterter is misspelled. Instead, use WithNullAlerter.
+	WithNullAlterter  = option.WithNullAlerter
```
jaypipes, line 39: I actually don't think we need to keep these around. Let's just fix the misspelled ones.

jaypipes (changes requested):
@GreyXor thanks very much for your contribution! If you could please sign your commit (do `git commit --amend -s`) and remove the misspelled aliases, that would be great. Thanks!

PR #278, opened 2021-09-21
context: allow reuse in auxiliary packages

fromanirh:
The PR #236 fixes a snapshot directory leakage issue, but also highlight a context lifecycle/ownership issue.

In general in ghw each package creates its own context, including the cases on which a package is consumed by another, for example `topology` and `pci`.
However, in this case, it would make more sense to reuse the context from the parent package.

Before the introduction of the the transparent snapshot support (#202), the above mechanism worked,
because each context, each time, was consuming over and over again the same parameters (e.g. environment variables); besides some resource waste, this had no negative effect.

When introducing snapshots in the picture, repeatedly unpacking the same snapshot to consume the same data is much more wasteful. So it makes sense now to introduce explicitly the concept of dependent context.

To move forward and make the ownership more explicit, we add a new function `NewWithContext` in all the subpackages.
The function `NewWithContext` will assume the context it gets as argument is ready to be consumed and it will use as-is, attempting no setup or teardown in any circumstances.

The usage rules are pretty simple:
1. In your client code, you most likely just need to use `New` as usual.
2. When consuming a package from another (e.g. `topology` from `pci`),    always use the `NewWithContext` function to get a handle for  the auxiliary package, passing through the context from the top-level  package.
3. Within a `NewWithContext` function, all calls to other   `NewWithContext` functions should be made to preserve the properties.

### DIFF
--- a/pkg/baseboard/baseboard.go
+++ b/pkg/baseboard/baseboard.go
@@ -65,6 +65,18 @@ func New(opts ...*option.Option) (*Info, error) {
 	return info, nil
 }
 
+// New returns a pointer to an Info struct containing information about the
+// host's baseboard, reusing a given context.
+// Use this function when you want to consume this package from another,
+// ensuring the two see a coherent set of resources.
+func NewWithContext(ctx *context.Context) (*Info, error) {
+	info := &Info{ctx: ctx}
+	if err := info.load(); err != nil {
+		return nil, err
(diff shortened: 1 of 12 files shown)


fromanirh:
replaced by https://github.com/jaypipes/ghw/pull/280

PR #280, opened 2021-09-27
allow passing in existing Context as an Option

jaypipes:
This patch simplifies some internal-only code that was added to ghw to
prevent re-creating Context objects inefficiently during snapshot
creation. We add a `context.WithContext` function that returns an
`option.Option` containing a pre-existing `Context` struct pointer,
embedded in the `Option` as a raw `interface{}` to get around package
recursive import issues.

Now, when packages like `pkg/gpu` or `pkg/pci` need to load the
`pkg/topology` package, they call
`topology.New(context.WithContext(ctx))`, passing in the
already-existing `Context` object. This preserves the original `New`
function signature without adding a new `NewWithContext` function
signature to all packages.

Alternative to #278

Signed-off-by: Jay Pipes <jaypipes@gmail.com>

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -11,8 +11,8 @@ import (
 	"github.com/jaypipes/ghw/pkg/snapshot"
 )
 
-// Concrete merged set of configuration switches that act as an execution
-// context when calling internal discovery methods
+// Context contains the merged set of configuration switches that act as an
+// execution context when calling internal discovery methods
 type Context struct {
 	Chroot               string
 	EnableTools          bool
@@ -24,10 +24,37 @@ type Context struct {
 	alert                option.Alerter
 }
 
+// WithContext returns an option.Option that contains a pre-existing Context
+// struct. This is useful for some internal code that sets up snapshots.
+func WithContext(ctx *Context) *option.Option {
+	return &option.Option{
+		Context: ctx,
+	}
+}
+
+// Exists returns true if the supplied (merged) Option already contains
+// a context.
+//
+// TODO(jaypipes): We can get rid of this when we combine the option and
+// context packages, which will make it easier to detect the presence of a
+// pre-setup Context.
+func Exists(opt *option.Option) bool {
+	return opt != nil && opt.Context != nil
+}
+
 // New returns a Context struct pointer that has had various options set on it
 func New(opts ...*option.Option) *Context {
 	merged := option.Merge(opts...)
-	ctx := &Context{
+	var ctx *Context
+	if merged.Context != nil {
+		var castOK bool
+		ctx, castOK = merged.Context.(*Context)
+		if !castOK {
+			panic("passed in a non-Context for the WithContext() function!")
+		}
+		return ctx
+	}
+	ctx = &Context{
 		alert:  option.EnvOrDefaultAlerter(),
 		Chroot: *merged.Chroot,
 	}
--- a/pkg/option/option.go
+++ b/pkg/option/option.go
@@ -133,6 +133,11 @@ type Option struct {
 	// PathOverrides optionally allows to override the default paths ghw uses internally
 	// to learn about the system resources.
 	PathOverrides PathOverrides
+
+	// Context may contain a pointer to a `Context` struct that is constructed
+	// during a call to the `context.WithContext` function. Only used internally.
+	// This is an interface to get around recursive package import issues.
+	Context interface{}
 }
 
 // SnapshotOptions contains options for handling of ghw snapshots
@@ -202,6 +207,8 @@ func WithPathOverrides(overrides PathOverrides) *Option {
 // a debug/troubleshoot aid more something users wants to do regularly.
 // Hence we allow that only via the environment variable for the time being.
 
+// Merge accepts one or more Options and merges them together, returning the
+// merged Option
 func Merge(opts ...*Option) *Option {
 	merged := &Option{}
 	for _, opt := range opts {
@@ -220,6 +227,9 @@ func Merge(opts ...*Option) *Option {
 		// intentionally only programmatically
 		if opt.PathOverrides != nil {
 			merged.PathOverrides = opt.PathOverrides
+		}
+		if opt.Context != nil {
+			merged.Context = opt.Context
 		}
 	}
 	// Set the default value if missing from mergeOpts
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -155,27 +155,29 @@ func (i *Info) String() string {
 // New returns a pointer to an Info struct that contains information about the
 // PCI devices on the host system
 func New(opts ...*option.Option) (*Info, error) {
-	return NewWithContext(context.New(opts...))
-}
-
-// NewWithContext returns a pointer to an Info struct that contains information about
-// the PCI devices on the host system. Use this function when you want to consume
-// the topology package from another package (e.g. gpu)
-func NewWithContext(ctx *context.Context) (*Info, error) {
+	merged := option.Merge(opts...)
+	ctx := context.New(merged)
 	// by default we don't report NUMA information;
 	// we will only if are sure we are running on NUMA architecture
-	arch := topology.ARCHITECTURE_SMP
-	topo, err := topology.NewWithContext(ctx)
-	if err == nil {
-		arch = topo.Architecture
-	} else {
-		ctx.Warn("error detecting system topology: %v", err)
-	}
 	info := &Info{
-		arch: arch,
+		arch: topology.ARCHITECTURE_SMP,
 		ctx:  ctx,
 	}
-	if err := ctx.Do(info.load); err != nil {
+	// we do this trick because we need to make sure ctx.Setup() gets
+	// a chance to run before any subordinate package is created reusing
+	// our context.
+	if err := ctx.Do(func() error {
+		topo, err := topology.New(context.WithContext(ctx))
+		if err == nil {
+			info.arch = topo.Architecture
+		} else {
+			ctx.Warn("error detecting system topology: %v", err)
+		}
+		if context.Exists(merged) {
+			return info.load()
+		}
+		return ctx.Do(info.load)
+	}); err != nil {
 		return nil, err
 	}
 	return info, nil
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -163,21 +163,27 @@ func New(opts ...*option.Option) (*Info, error) {
 		arch: topology.ARCHITECTURE_SMP,
 		ctx:  ctx,
 	}
+
 	// we do this trick because we need to make sure ctx.Setup() gets
 	// a chance to run before any subordinate package is created reusing
 	// our context.
-	if err := ctx.Do(func() error {
+	loadDetectingTopology := func() error {
 		topo, err := topology.New(context.WithContext(ctx))
 		if err == nil {
 			info.arch = topo.Architecture
 		} else {
 			ctx.Warn("error detecting system topology: %v", err)
 		}
-		if context.Exists(merged) {
-			return info.load()
-		}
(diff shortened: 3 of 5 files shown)


pkg/context/context.go
```diff
@@ -24,10 +24,27 @@ type Context struct {
...
 // New returns a Context struct pointer that has had various options set on it
 func New(opts ...*option.Option) *Context {
 	merged := option.Merge(opts...)
-	ctx := &Context{
+	var ctx *Context
+	if merged.Context != nil {
+		var castOK bool
+		ctx, castOK = merged.Context.(*Context)
+		if !castOK {
+			panic("passed in a non-Context for the WithContext() function!")
```
fromanirh, line 43: I usually dislike `panic`ing into library code, but this is probably one of the very few places on which is useful. So I'm fine.

pkg/option/option.go
```diff
@@ -133,6 +133,11 @@ type Option struct {
 	// PathOverrides optionally allows to override the default paths ghw uses internally
 	// to learn about the system resources.
 	PathOverrides PathOverrides
+
+	// Context may contain a pointer to a `Context` struct that is constructed
+	// during a call to the `context.WithContext` function. Only used internally.
+	// This is an interface to get around recursive package import issues.
+	Context interface{}
```
fromanirh, line 140: While I totally see why you are doing this, this is a smell. There are options (no pun intended) to fix this, but all of them will require a overhaul of how we handle `context` and `option`, which must be either a blocker for 1.0.0 or deferred post 1.0.0 when we can break compatibility.

pkg/pci/pci.go
```diff
@@ -155,17 +155,11 @@ func (i *Info) String() string {
...
 func New(opts ...*option.Option) (*Info, error) {
-	return NewWithContext(context.New(opts...))
-}
-
-// NewWithContext returns a pointer to an Info struct that contains information about
-// the PCI devices on the host system. Use this function when you want to consume
-// the topology package from another package (e.g. gpu)
-func NewWithContext(ctx *context.Context) (*Info, error) {
+	ctx := context.New(opts...)
+	topo, err := topology.New(context.WithContext(ctx))
```
fromanirh, line 159: from the consumer perspective the API looks nicer and more coherent with the rest of `ghw`, which is great. However here you'll still need to use a nested function (or some other trick) like I'm doing into
https://github.com/jaypipes/ghw/pull/278/commits/691659d5a0f68705b4ca7f3b4e285d22ad7a6cf2#diff-a6b7c579eae86e07718647003b278fce2940fccd215fa7db50f05f697691847fR158
to make sure that `topology.New` only runs *after* `pci` set the snapshot (if needed)

fromanirh (approved):
Comparing this PR with my https://github.com/jaypipes/ghw/pull/278 is a quite hard call. The things I don't like in this PR are the awkward usage of `interface{}` into `option.Option`, and the type check + panic in `context.Context`. It is pretty much to avoid them that I introduced the clumsy `NewWithContext` API.
This PR however is much more self contained, and treats a specific, uncommon flow as such instead of sprinking changes all over the codebase to support that flow.

From the project as a whole perspective, I think this PR (bar the comment inside about the interaction between `topology` and `pci`) is a tad better because it is indeed more self contained and in general follows more the existing direction and feeling of the ghw API.

Furthermore, we're not adding new APIs here (opposite to what I was doing). This is more and more relevant as 1.0.0 approaches. Should we grow not too happy about the approach implemented here, we can add APIs later; OTOH removing them is always harder.

Last but not least, I think that we should -in due time- just take the chance to revisit the way the `option` and `context` package interact and maybe redesign them. This is easier done going with this approach (minimal change) than with mine (more API surface).
Yet another reason to prefer this change over https://github.com/jaypipes/ghw/pull/278.

Not merging because of inline comment about `pci` and `topology`.

jaypipes:
> Comparing this PR with my #278 is a quite hard call. The things I don't like in this PR are the awkward usage of `interface{}` into `option.Option`, and the type check + panic in `context.Context`. It is pretty much to avoid them that I introduced the clumsy `NewWithContext` API. This PR however is much more self contained, and treats a specific, uncommon flow as such instead of sprinking changes all over the codebase to support that flow.

Yeah, I totally hear you. It's a bit of choosing between two not-great options, I agree.

> From the whole project perspective, I think this PR (bar the comment inside about the interaction between `topology` and `pci`) is a tad better because it is indeed more self contained and in general follows more the existing direction and feeling of the ghw API.
> 
> Furthermore, we're not adding new APIs here (opposite to what I was doing). This is more and more relevant as 1.0.0 approaches. Should we grow not too happy about the approach implemented here, we can add APIs later; OTOH removing them is always harder.

Yes, agreed.

> Last but not least, I think that we should -in due time- just take the chance to revisit the way the `option` and `context` package interact and maybe redesign them. This is easier done going with this approach (minimal change) than with mine (more API surface). Yet another reason to prefer this change over #278.

We should combine the `context` and `option` packages, yes. And I plan to do that before 1.0.

pkg/context/context.go
jaypipes: yeah, this is more of a compile-time thing since mostly it's just checking to make sure we're setting the interface properly.

pkg/option/option.go
jaypipes: agreed. I plan to merge option and context packages before 1.0

fromanirh: Awesome. I'd love to help here.

pkg/pci/pci.go
jaypipes: ah, I see now. OK, fixing up.

fromanirh:
> > From the whole project perspective, I think this PR (bar the comment inside about the interaction between `topology` and `pci`) is a tad better because it is indeed more self contained and in general follows more the existing direction and feeling of the ghw API.
> > Furthermore, we're not adding new APIs here (opposite to what I was doing). This is more and more relevant as 1.0.0 approaches. Should we grow not too happy about the approach implemented here, we can add APIs later; OTOH removing them is always harder.
> 
> Yes, agreed.

Indeed we agree, so we have a way forward! Let's pick this PR over https://github.com/jaypipes/ghw/pull/278.The only missing item is how we want to handle https://github.com/jaypipes/ghw/commit/691659d5a0f68705b4ca7f3b4e285d22ad7a6cf2#diff-a6b7c579eae86e07718647003b278fce2940fccd215fa7db50f05f697691847fR158 - I'm fine with any approach from changing this PR to a followup one. Once this is settled I'll close my #278

> > Last but not least, I think that we should -in due time- just take the chance to revisit the way the `option` and `context` package interact and maybe redesign them. This is easier done going with this approach (minimal change) than with mine (more API surface). Yet another reason to prefer this change over #278.
> 
> We should combine the `context` and `option` packages, yes. And I plan to do that before 1.0.

Awesome news!

pkg/pci/pci.go
fromanirh: So I overlooked a key detail. Apologies! if you just add the trick I mentioned and run the tests, you'll get something like
```
-- FAIL: TestPCIDeviceRevision (0.22s)
    --- FAIL: TestPCIDeviceRevision/0000:07:03.0 (0.00s)
        pci_linux_test.go:86: got nil device for address "0000:07:03.0"
    --- FAIL: TestPCIDeviceRevision/0000:05:00.0 (0.00s)
        pci_linux_test.go:86: got nil device for address "0000:05:00.0"
setup(0xc00015e070) -> root="/tmp/ghw-snapshot-383180324"
setup(0xc00015e070) -> root="/tmp/ghw-snapshot-217854259"
teardown(0xc00015e070) -> root="/tmp/ghw-snapshot-217854259"
WARNING: failed to read /sys/bus/pci/devices
teardown(0xc00015e070) -> root="/tmp/ghw-snapshot-217854259"
```
(extra traces added by myself).

In the following of this post:
1. "outer package" = `pci`
2. "inner package" = `topology`

This means that this PR solves the problem of passing the context to subordinate packages, but still we get the snapshot handling incorrect:
1. outer package does setup
2. inner package does setup (again, wasteful but correct)
2. innerr package does their load
3. outer package *does teardown* - now the snapshot is gone but the state is inconsistent: the context passed through still believes the snapshot is availaible
4. outer package does their load - and this will fail!
5. outer package does teardown - which is a NOP at this point

IOW, the nesting of operations looks like
```
outer setup
   inner setup
     inner load
   inner teardown
   outer load
outer teardown
```

what we actually need is *only the outer package to do the setup/teardown* so we need a way to convey the inner packages to skip any setup/teardown. Just consume the context as is. It should look like
```
outer setup
    inner load
  outer load
outer teardown
```

So we need a way to convey the inner package "leave the context alone, just consume it". A minimal, and unfortunately a little uglier way, to do this could be:
1. in `context.go` add ```
2.

jaypipes: Good detective work, @fromanirh :) I'll apply your patch and repush!

pkg/pci/pci.go
```diff
@@ -155,27 +155,29 @@ func (i *Info) String() string {
...
+		if err == nil {
+			info.arch = topo.Architecture
+		} else {
+			ctx.Warn("error detecting system topology: %v", err)
+		}
+		if context.Exists(merged) {
+			return info.load()
+		}
+		return ctx.Do(info.load)
+	}); err != nil {
```
fromanirh, line 180: I'm deeply sorry but I managed to fumble this up. Now the order of operation is correct but we do things twice - the culprit is `ctx.Do` at the end. We need to reshuffle a bit the components: `ctx.Do`, anonymous func and `context.Exists`. This patch should fix:
```
diff --git a/pkg/pci/pci.go b/pkg/pci/pci.go
index 68d51b0..7411277 100644
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -163,21 +163,27 @@ func New(opts ...*option.Option) (*Info, error) {
                arch: topology.ARCHITECTURE_SMP,
                ctx:  ctx,
        }
+
        // we do this trick because we need to make sure ctx.Setup() gets
        // a chance to run before any subordinate package is created reusing
        // our context.
-       if err := ctx.Do(func() error {
+       loadDetectingTopology := func() error {
                topo, err := topology.New(context.WithContext(ctx))
                if err == nil {
                        info.arch = topo.Architecture
                } else {
                        ctx.Warn("error detecting system topology: %v", err)
                }
-               if context.Exists(merged) {
-                       return info.load()
-               }
-               return ctx.Do(info.load)
-       }); err != nil {
+               return info.load()
+       }
+
+       var err error
+       if context.Exists(merged) {
+               err = loadDetectingTopology()
+       } else {
+               err = ctx.Do(loadDetectingTopology)
+       }
+       if err != nil {
                return nil, err
        }
        return info, nil
```
IOW the function should read like
```
func New(opts ...*option.Option) (*Info, error) {
	merged := option.Merge(opts...)
	ctx := context.New(merged)
	// by default we don't report NUMA information;
	// we will only if are sure we are running on NUMA architecture
	info := &Info{
		arch: topology.ARCHITECTURE_SMP,
		ctx:  ctx,
	}

	// we do this trick because we need to make sure ctx.Setup() gets
	// a chance to run before any subordinate package is created reusing
	// our context.
	loadDetectingTopology := func() error {
		topo, err := topology.New(context.WithContext(ctx))
		if err == nil {
			info.arch = topo.Architecture
		} else {
			ctx.Warn("error detecting system topology: %v", err)
		}
		return info.load()
	}

	var err error
	if context.Exists(merged) {
		err = loadDetectingTopology()
	} else {
		err = ctx.Do(loadDetectingTopology)
	}
	if err != nil {
		return nil, err
	}
	return info, nil
}
```

And I verified carefully this version and looks really good this time, so third time should be the charm!

fromanirh: And I'm thinking about a test to catch this. I'll add in a followup PR.

PR #281, opened 2021-09-29
context: detect and report conflicting options

ffromani:
`WithChroot` and `WithSnapshot` conflicts to each other, and the users can get surprising results if they try to use them
at the same time.

Now: chroot is expected to be much more popular than snapshots, being the latter useful mostly for tests or offline debug/troubleshoot, but still the code should detect and handle these conflicts.

This is made a bit awkward because the current `context` pkg API. Once we are free to break compatibility, we can make the flow more linear.

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -7,6 +7,8 @@
 package context
 
 import (
+	"fmt"
+
 	"github.com/jaypipes/ghw/pkg/option"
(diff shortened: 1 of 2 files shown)


fromanirh:
> Thanks @fromanirh! Sorry for the delay in merging this!

It's fine! thank you for reviewing!

PR #286, opened 2021-11-03
pci: don't pass chroot to pcidb with snapshots

fromanirh:
The snapshot implementation leverages chroot and transparently set it to the unpacked snapshot root. This is the intentional behaviour and ghw supports it.
But the `pci` package passes throu the chroot settings to `pcidb` and the latter doesn't know or care about ghw snapshots - and rightly so!

Hence, we need to pass through the `chroot` setting only if explicitely set by client code - so if no snapshot settings are present.


jaypipes (approved):
Yeah, nice catch @fromanirh! :+1:

PR #287, opened 2021-11-11
Update block_windows.go

adumandix:
issue -> type *uint32 is a pointer. It would not equals forever.


jaypipes:
Sorry @adumandix I don't know what this PR is supposed to be doing... plus, it changes all import paths to reference your personal GH repo location for some reason.

Issue #289, opened 2021-11-22
update golang stack in CI


fromanirh:
Currently we support golang 1.14, 1.15 and 1.16 on CI. Golang 1.17 was released quite some time ago and got already patch releases, and golang 1.18 is not too far ahead.
Let's rotate our supported versions and run golang 1.15, 1.16 and 1.17 on CI.
I expect new lint/fmt fixes coming with this update.

jaypipes:
Agreed, @fromanirh. I've been developing with 1.17 on my local workstation for a while now. Our [go.mod is currently at `1.12`](https://github.com/jaypipes/ghw/blob/004673db31bce4f54fa4f5525a9712d03710e999/go.mod#L3). Shall we up that to 1.15?

fromanirh:
> Agreed, @fromanirh. I've been developing with 1.17 on my local workstation for a while now. Our [go.mod is currently at `1.12`](https://github.com/jaypipes/ghw/blob/004673db31bce4f54fa4f5525a9712d03710e999/go.mod#L3). Shall we up that to 1.15?

Yes. I'do as part of this PR

PR #290, opened 2021-11-23
toolchain update

fromanirh:
Rotate the supported golang versions: golang 1.14, add golang 1.17. Includes fix to keep CI green.

Fixes: https://github.com/jaypipes/ghw/issues/289

### DIFF
--- a/pkg/block/block_linux_test.go
+++ b/pkg/block/block_linux_test.go
@@ -4,6 +4,7 @@
 // See the COPYING file in the root project directory for full text.
 //
 
+//go:build linux
 // +build linux
 
 package block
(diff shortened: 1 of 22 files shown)


pkg/block/block_linux_test.go
```diff
@@ -4,6 +4,7 @@
 // See the COPYING file in the root project directory for full text.
 //
 
+//go:build linux
 // +build linux
```
jaypipes, line 8: in a future PR, we should move this to the top of the file, to be consistent with all the other files. Same with other _test.go fles...

jaypipes (approved):
Awesome, thank you sir!

PR #299, opened 2022-01-31
util: add helper to avoid long format strings

fromanirh:
A relatively common pattern in the ghw codebase
is concatenating quite a lot of strings inside
a format string. This leads to a pattern like '%s%s%s%s%s'
which are hard to read and to debug.

Ad a simple helper to concatenate strings to simplify the
format strings.

This may be the definitive solution here, but
it should improve the maintenability and readability,
and it's very cheap to implement and maintain.

### DIFF
--- a/pkg/baseboard/baseboard.go
+++ b/pkg/baseboard/baseboard.go
@@ -7,8 +7,6 @@
 package baseboard
 
 import (
-	"fmt"
-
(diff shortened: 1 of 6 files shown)


jaypipes (approved):
A perfectly reasonable utility, thanks Francesco :)

Issue #297, opened 2022-01-31
Properly detect iSCSI disks


avishayt:
iSCSI block devices are currently detected as "HDD" - it would be beneficial to detect them as "iSCSI".

jaypipes:
@avishayt Sorry to get to this so late, but iSCSI is not the drive type, AFAIK. SCSI is the storage controller interface, and all iSCSI drives should appear as HDD drive type and SCSI storage controller... is there a reason we needed a separate iSCSI drive type?

@fromanirh I am considering reverting this PR...

PR #300, opened 2022-01-31
Detect iSCSI devices

avishayt:
Fixes: https://github.com/jaypipes/ghw/issues/297


pkg/block/block_linux.go
```diff
@@ -277,6 +277,9 @@ func disks(ctx *context.Context, paths *linuxpath.Paths) []*Disk {
 		size := diskSizeBytes(paths, dname)
 		pbs := diskPhysicalBlockSizeBytes(paths, dname)
 		busPath := diskBusPath(paths, dname)
+		if strings.Contains(busPath, "-iscsi-") {
+			driveType = DRIVE_TYPE_ISCSI
```
fromanirh, line 281: I wonder if there is another way to know if a device is iscsi or not

fromanirh (commented):
thanks for the PR! indeed acknowledging ISCSI devices would be nice.
It would be real cool to add unit test(s) to exercise this code, even if it's a simple change, to make it unlikely to break in the future

pkg/block/block_linux.go
avishayt: Not that I know of.  This method has worked consistently for me.

avishayt:
> thanks for the PR! indeed acknowledging ISCSI devices would be nice. It would be real cool to add unit test(s) to exercise this code, even if it's a simple change, to make it unlikely to break in the future

Added a unit test - thanks!

fromanirh:
thanks for the updates! codewise LGTM - we can further improve the testing in the future if needed/feasible.
The only thing left is the DCO check which is failing - could you PTAL to the instructions (https://github.com/jaypipes/ghw/pull/300/checks?check_run_id=5556037529) and see if there is a easy fix? feel free to get in touch if instructions are not clear

avishayt:
> thanks for the updates! codewise LGTM - we can further improve the testing in the future if needed/feasible. The only thing left is the DCO check which is failing - could you PTAL to the instructions (https://github.com/jaypipes/ghw/pull/300/checks?check_run_id=5556037529) and see if there is a easy fix? feel free to get in touch if instructions are not clear

Fixed, thanks for the pointer

pkg/block/block_linux.go
fromanirh: We can surely start like this

Issue #298, opened 2022-01-31
Properly detect dm devices


avishayt:
With multipath, we currently see that the drive type is "unknown".  It would be great to see it as "multipath" or even something more generic such as "mapper", and it would be even greater to see which other devices were associated with it.

fromanirh:
That's a good point indeed. These are abstractions provided by the linux kernel which happen to be presented using the same interface as the real devices. From this perspective it is out of scope from a hardware abstraction library.

Maybe it's too early, but it seems there it seems to be a trend emerging about features like DM and loopback which are not hardware proper but are traditionally modeled as block devicse and are similar enough to hardware to have a real use cases.

I'm really thinking aloud, but maybe a way forward could be worth to have something like a ghw companion/extension library to handle pseudo devices? I'll file a issue in the future to discuss this option.

jaypipes:
@avishayt @fromanirh I might be convinced to add an entirely separate Go package for devicemappers, though. Such a package would treat device mappings/mappers as a separate construct (that would reference `pkg/block.Disk` structs within it.

PR #301, opened 2022-01-31
Properly detect dm devices

avishayt:
Fixes: https://github.com/jaypipes/ghw/issues/298


jaypipes:
@avishayt I apologize for the delay in getting to these reviews, sir. Is there a reason you closed out your PRs?

avishayt:
Wasn't sure the project was still active :)
Reopening

fromanirh (commented):
thanks for the change! the code per se looks good! It would be real nice to have some tests to cover this change, in order to avoid future breakage.
It would be also real nice to have an example of the sysfs directory tree involved here.
Perhaps `ghw-snapshot` can help to capture that?

pkg/block/block.go
```diff
@@ -166,6 +169,7 @@ type Disk struct {
 	SerialNumber string       `json:"serial_number"`
 	WWN          string       `json:"wwn"`
 	Partitions   []*Partition `json:"partitions"`
+	Members      []string     `json:"members"`
```
fromanirh, line 172: LGTM but let's hear from @jaypipes because of the API-level change

pkg/block/block_linux.go
```diff
@@ -271,7 +284,7 @@ func disks(ctx *context.Context, paths *linuxpath.Paths) []*Disk {
 		driveType, storageController := diskTypes(dname)
 		// TODO(jaypipes): Move this into diskTypes() once abstracting
 		// diskIsRotational for ease of unit testing
-		if !diskIsRotational(ctx, paths, dname) {
+		if driveType != DRIVE_TYPE_MAPPER && !diskIsRotational(ctx, paths, dname) {
```
fromanirh, line 287: we will probably need to review this flow more deeply, but that's totally material for another PR. This change LGTM for this PR

fromanirh (approved):
LGTM, thanks for the updates.

Issue #199, opened 2020-09-18
Support for ARM


sivachandran:
Presently CPU info on ARM is empty(`cpu (0 physical packages, 0 cores, 0 hardware threads)`) as there are many fields missing in ARM's `/proc/cpuinfo`.

**Sample ARM /proc/cpuinfo**
```
processor	: 0
BogoMIPS	: 108.00
Features	: fp asimd evtstrm crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd08
CPU revision	: 3

processor	: 1
BogoMIPS	: 108.00
Features	: fp asimd evtstrm crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd08
CPU revision	: 3

processor	: 2
BogoMIPS	: 108.00
Features	: fp asimd evtstrm crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd08
CPU revision	: 3

processor	: 3
BogoMIPS	: 108.00
Features	: fp asimd evtstrm crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd08
CPU revision	: 3

Hardware	: BCM2835
Revision	: c03111
Serial		: 100000002f8e83b3
Model		: Raspberry Pi 4 Model B Rev 1.1
```

More samples can be found here https://github.com/shirou/gopsutil/issues/881

Even if we change the implementation to ignore missing fields, the values are of available fields are numeric. They need to be converted into human readable string for meaning usage.

Seems the right way to do this is to use combination of `lscpu` and `/proc/cpuinfo` results. As described here http://suihkulokki.blogspot.com/2018/02/making-sense-of-proccpuinfo-on-arm.html, the `lscpu` is actively maintained and provides reliable information. On my ARM platform(RaspberryPi) I can confirm the presence of `lscpu`. I believe it would be available in other ARM platforms as well.

Is it okay to add support for ARM with help of of both `lscpu` and `/proc/cpuinfo`?

jaypipes:
> Is it okay to add support for ARM with help of of both `lscpu` and `/proc/cpuinfo`?

Hi @sivachandran! Yes, absolutely it would be awesome to add support for ARM. I don't have time to work on this at the moment, though. Are you interested in contributing here?

[...]

PR #303, opened 2022-03-07
Add aarch64 and RISCV64 support

rockrush:
Depend on sysfs instead of /proc/cpuinfo, for content of this file varies
across CPU architectures, and doesn't offer unified info tag.

Solves issue #199 .

Signed-off-by: Xingyou Chen <rockrush@rockwork.org>

### DIFF
--- a/cmd/ghwc/commands/cpu.go
+++ b/cmd/ghwc/commands/cpu.go
@@ -43,12 +43,12 @@ func showCPU(cmd *cobra.Command, args []string) error {
 				// pretty-print the (large) block of capability strings into rows
 				// of 6 capability strings
 				rows := int(math.Ceil(float64(len(proc.Capabilities)) / float64(6)))
-				for row := 1; row < rows; row = row + 1 {
-					rowStart := (row * 6) - 1
+				for row := 0; row < rows; row = row + 1 {
+					rowStart := (row * 6)
 					rowEnd := int(math.Min(float64(rowStart+6), float64(len(proc.Capabilities))))
 					rowElems := proc.Capabilities[rowStart:rowEnd]
 					capStr := strings.Join(rowElems, " ")
-					if row == 1 {
+					if row == 0 {
 						fmt.Printf("  capabilities: [%s\n", capStr)
 					} else if rowEnd < len(proc.Capabilities) {
 						fmt.Printf("                 %s\n", capStr)
--- a/pkg/cpu/cpu_linux.go
+++ b/pkg/cpu/cpu_linux.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -32,6 +33,24 @@ func (i *Info) load() error {
 	return nil
 }
 
+func ProcByID(procs []*Processor, id int) *Processor {
+	for pid := range procs {
+		if procs[pid].ID == id {
+			return procs[pid]
+		}
+	}
+	return nil
+}
+
+func CoreByID(cores []*ProcessorCore, id int) *ProcessorCore {
+	for cid := range cores {
+		if cores[cid].Index == id {
+			return cores[cid]
+		}
+	}
+	return nil
+}
+
 func processorsGet(ctx *context.Context) []*Processor {
 	procs := make([]*Processor, 0)
 	paths := linuxpath.New(ctx)
@@ -46,6 +65,7 @@ func processorsGet(ctx *context.Context) []*Processor {
 	procAttrs := make([]map[string]string, 0)
 	curProcAttrs := make(map[string]string)
 
+	// Parse /proc/cpuinfo
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
 		line := strings.TrimSpace(scanner.Text())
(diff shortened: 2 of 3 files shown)


fromanirh:
Thanks for this PR!
I'm generally in favour of moving from `/proc/cpuinfo` to `sysfs` to learn about the CPU data.
Let's check this change is covered by the existing tests, and let's add more of them if needed
other than that, at glance looks good! I'll have a deeper look later

cmd/ghwc/commands/cpu.go
```diff
@@ -43,12 +43,12 @@ func showCPU(cmd *cobra.Command, args []string) error {
 				// pretty-print the (large) block of capability strings into rows
 				// of 6 capability strings
 				rows := int(math.Ceil(float64(len(proc.Capabilities)) / float64(6)))
-				for row := 1; row < rows; row = row + 1 {
-					rowStart := (row * 6) - 1
+				for row := 0; row < rows; row = row + 1 {
```
fromanirh, line 46: this change seems to have some merit, but it should be moved on its own commit (or possibly on its own PR) since it seems more refactoring/cleanup than actually part of the core concept of this PR

pkg/cpu/cpu_linux.go
```diff
@@ -64,85 +84,69 @@ func processorsGet(ctx *context.Context) []*Processor {
...
 	}
 
-	// Build a set of physical processor IDs which represent the physical
-	// package of the CPU
-	setPhysicalIDs := make(map[int]bool)
-	for _, attrs := range procAttrs {
-		pid, err := strconv.Atoi(attrs["physical id"])
-		if err != nil {
+	// Iterate on /sys/devices/system/cpu/cpuN, not on /proc/cpuinfo
+	Entries, _ := ioutil.ReadDir(paths.SysDevicesSystemCPU)
```
fromanirh, line 88: even if (very) unlikely, should we probably handle the error here

pkg/cpu/cpu_linux.go
```diff
@@ -64,85 +84,69 @@ func processorsGet(ctx *context.Context) []*Processor {
...
-		p.Model = first["model name"]
-		p.Vendor = first["vendor_id"]
-		numCores, err := strconv.Atoi(first["cpu cores"])
-		if err != nil {
-			continue
-		}
-		p.NumCores = uint32(numCores)
-		numThreads, err := strconv.Atoi(first["siblings"])
-		if err != nil {
+		lcoreID, error := strconv.Atoi(lcore.Name()[3:]) // get "N" out of "cpuN"
```
fromanirh, line 95: (very) minor: since we are already using a regexp, we can evaluate if we can use a group to extract the cpuid right away instead of doing in two stages. Not blocking, can be done in a later PR/commit

pkg/cpu/cpu_linux.go
```diff
@@ -64,85 +84,69 @@ func processorsGet(ctx *context.Context) []*Processor {
...
-				continue
-			}
-			coreID, err := strconv.Atoi(procAttrs[lpidx]["core id"])
-			if err != nil {
-				continue
+		// Fetch CPU ID
+		physIdPath := fmt.Sprintf(
+			"%s/cpu%d/topology/physical_package_id",
+			paths.SysDevicesSystemCPU,
+			lcoreID,
```
fromanirh, line 104: even if this is only relevant for linux, I prefer to use `filepath.Join` when building paths.

pkg/cpu/cpu_linux.go
```diff
@@ -64,85 +84,69 @@ func processorsGet(ctx *context.Context) []*Processor {
...
-			lpid, err := strconv.Atoi(procAttrs[lpidx]["processor"])
-			if err != nil {
-				continue
-			}
-			coreID, err := strconv.Atoi(procAttrs[lpidx]["core id"])
-			if err != nil {
-				continue
+		// Fetch CPU ID
+		physIdPath := fmt.Sprintf(
+			"%s/cpu%d/topology/physical_package_id",
```
fromanirh, line 102: could you please add a comment to clarify why this is the right file to learn about the cpu id in the sysfs topology? https://www.kernel.org/doc/html/v5.5/admin-guide/cputopology.html

pkg/cpu/cpu_linux.go
```diff
@@ -64,85 +84,69 @@ func processorsGet(ctx *context.Context) []*Processor {
...
-				cores = append(cores, core)
+			if len(procAttrs[lcoreID]["vendor_id"]) != 0 {
+				proc.Vendor = procAttrs[lcoreID]["vendor_id"]
+			} else if len(procAttrs[lcoreID]["isa"]) != 0 { // RISCV64
+				proc.Vendor = procAttrs[lcoreID]["isa"]
 			}
+			procs = append(procs, proc)
+		}
+
+		// Fetch Core ID
```
fromanirh, line 133: same comment about building paths

fromanirh (commented):
initial review
first of all, thanks for working in this area. I think this is a welcome change.
To move forward, there are a couple of things to address:
1. we would need some aarch64 and riscv64 machines to run at least some CI on; unfortunately, code outside CI bitrots pretty fast. I'm not aware of option (not sure gh actions run on arm, IIRC not)
2. it would be very useful to have some docs highligthing how the topology exposed by sysfs exports data in case of x86_64 and risc64 and others. E.g., what's a "core" for riscv64? is it the same thing we call "core" in x86_64?

Itxaka:
FYI you can run tests on qemu arm64 under github ci. Yes, its slow as hell, and you need to ssh into the VM to run your tests, but seems to work as expected, we run our battery of tests on it, even installing and restarting the vm to test different things and we can say it works.

See a vagrantfile to bring machine X up on qemu-system-aarch64: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/tests/Vagrantfile#L30 and the makefile that brings it up: https://github.com/rancher-sandbox/cOS-toolkit/blob/master/make/Makefile.test#L61

Let me know if you need any help in this area (testing on aarch64 vms on github ci!) :)

jak3kaj:
This PR resolves CPU discovery issues on an ARMv7 network switch as well.

I would like to see this PR merged into main.

fromanirh:
adding arm support would be very nice indeed. Unfortunately we still need to sort out the CI and the supportability aspects, because we very much want to avoid the feature to break (or to break silently!) when the projects moves on.

jak3kaj:
Understood.  

I'm going to work on adding arm support for some of the data that's usually stored in dmi on x86.  I will also work on setting up some arm CI in a fork and I'll circle back to see if that will be suitable for this project too.

glimchb:
FYI @jaypipes merged #333 , this can be closed

Itxaka:
nice!

PR #304, opened 2022-03-17
Get partition labels from udev

Itxaka:
This mainly reuses the udevInfo method to obtain the udev info for a
given partition in a given disk instead of having the label value always
empty.

This separates the udevInfo method into 2 distinct methods, one for disk
and one for partition.

This also opens the possibility of obtaining more info for partitions,
i.e. the FS of an unmounted partition.

Signed-off-by: Itxaka <igarcia@suse.com>

PR #305, opened 2022-03-18
Get partition labels from udev

Itxaka:
This mainly reuses the udevInfo method to obtain the udev info for a
given partition in a given disk instead of having the label value always
empty.

This separates the udevInfo method into 2 distinct methods, one for disk
and one for partition.

This also opens the possibility of obtaining more info for partitions,
i.e. the FS of an unmounted partition.

This requires https://github.com/jaypipes/ghw/pull/304 as that brings the udevInfoPartition


Itxaka:
will rebase if #304 gets in :)

Itxaka:
rebased!

pkg/block/block_linux_test.go
```diff
@@ -251,3 +251,34 @@ func TestDiskPartLabel(t *testing.T) {
...
+	_ = ioutil.WriteFile(filepath.Join(paths.RunUdevData, "b259:0"), []byte(fmt.Sprintf("E:ID_FS_TYPE=%s\n", fsType)), 0644)
+	fs := diskPartTypeUdev(paths, "sda", "sda1")
+	if fs != fsType {
+		t.Fatalf("Got fs %s but expected %s", fs, fsType)
+	}
+
+	// Check empty fs if not found
+	fs = diskPartTypeUdev(paths, "sda", "sda2")
+	if fs != util.UNKNOWN {
+		t.Fatalf("Got label %s, but expected %s label", fs, util.UNKNOWN)
```
jaypipes, line 282: s/label/partition type/g :)

pkg/block/block_linux.go
```diff
@@ -237,6 +240,18 @@ func diskPartLabel(paths *linuxpath.Paths, disk string, partition string) string
 	return util.UNKNOWN
 }
 
+func diskPartTypeUdev(paths *linuxpath.Paths, disk string, partition string) string {
```
jaypipes, line 243: Please add a quick docstring indicating this function is used as a fallback when `partitionInfo()` fails to determine a partition type, which happens when the partition is not mounted.

pkg/block/block_linux_test.go
```diff
@@ -251,3 +251,34 @@ func TestDiskPartLabel(t *testing.T) {
...
+
+	_ = os.MkdirAll(paths.SysBlock, 0755)
+	_ = os.MkdirAll(paths.RunUdevData, 0755)
+
+	// Emulate a disk with one partition with label TEST_LABEL_GHW
+	_ = os.Mkdir(filepath.Join(paths.SysBlock, "sda"), 0755)
+	_ = os.Mkdir(filepath.Join(paths.SysBlock, "sda", "sda1"), 0755)
+	_ = ioutil.WriteFile(filepath.Join(paths.SysBlock, "sda", "sda1", "dev"), []byte("259:0\n"), 0644)
+	_ = ioutil.WriteFile(filepath.Join(paths.RunUdevData, "b259:0"), []byte(fmt.Sprintf("E:ID_FS_TYPE=%s\n", fsType)), 0644)
+	fs := diskPartTypeUdev(paths, "sda", "sda1")
```
jaypipes, line 274: consider renaming `fsType` to `expectedPartType` and `fs` variables to `pt` to align with the "partition type" terminology (and yes, I realize udev uses "FS_TYPE" term :) )

Itxaka:
@fromanirh any updates on this? thanks!

jaypipes (changes requested):
@Itxaka awesome stuff, thank you so much! Just a few little nits inline to address, otherwise looks great :)

pkg/block/block_linux_test.go
Itxaka: done!

pkg/block/block_linux_test.go
Itxaka: done!

fromanirh:
I see @jaypipes is already on it; I will have a look myself today/tomorrow as well

jaypipes (approved):
rock on :) thanks @Itxaka!

PR #309, opened 2022-03-19
Update core dependencies, Dockerfile, fix error in run-against-snapshot

taigrr:
While trying to validate a snapshot created for #293 , I ran into issues using the run-against-snapshot script. It seems some of the tags for the dependencies were removed upstream, preventing the container from building.

Additionally, there was a bug in the script preventing it from running without an environment variable being set.

Finally, stretch has been unsupported for over a year at this point, and the CA store was causing issues downloading from package registries, so I updated the base image to buster, and the go tag to 1.15 from 1.13 to match the go.mod


fromanirh:
thanks for this change! at glance LGTM, I'll have a proper review asap

fromanirh:
@jaypipes was fast, I was about to approve :)

PR #310, opened 2022-03-21
Add WithLoopDevices to report loop devices

Itxaka:
Currently we skipp over all loop devices in the system and with no
possibility of reporting those.

This patch introduces a new option to Block that allows us to report
loop devices as disk and their partitions like if they were normal
disks.

This patch does not modify any existing functionality or alters anything
with the default values.

With WithLoopDevices enabled this reports the same info for loop devices
as with other disks, including size, fs, mountpoints and so on.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -29,7 +29,7 @@ const (
 	DRIVE_TYPE_FDD               // Floppy disk drive
 	DRIVE_TYPE_ODD               // Optical disk drive
 	DRIVE_TYPE_SSD               // Solid-state drive
-	DRIVE_TYPE_LOOP              // loop device
+	DRIVE_TYPE_VIRTUAL           // virtual drive i.e. loop devices
 )
 
 var (
@@ -52,7 +52,7 @@ var (
 		"fdd":     DRIVE_TYPE_FDD,
 		"odd":     DRIVE_TYPE_ODD,
 		"ssd":     DRIVE_TYPE_SSD,
-		"loop":    DRIVE_TYPE_LOOP,
+		"virtual": DRIVE_TYPE_VIRTUAL,
 	}
 )
 
@@ -95,7 +95,7 @@ const (
 	STORAGE_CONTROLLER_NVME                      // Non-volatile Memory Express
 	STORAGE_CONTROLLER_VIRTIO                    // Virtualized storage controller/driver
 	STORAGE_CONTROLLER_MMC                       // Multi-media controller (used for mobile phone storage devices)
-	STORAGE_CONTROLLER_LOOP                      // loop device
+	STORAGE_CONTROLLER_VIRTUAL                   // virtual controller, i.e. for a loop device
 )
 
 var (
@@ -106,7 +106,7 @@ var (
 		STORAGE_CONTROLLER_NVME:    "NVMe",
 		STORAGE_CONTROLLER_VIRTIO:  "virtio",
 		STORAGE_CONTROLLER_MMC:     "MMC",
-		STORAGE_CONTROLLER_LOOP:    "loop",
+		STORAGE_CONTROLLER_VIRTUAL: "virtual",
 	}
 
 	// NOTE(fromani): the keys are all lowercase and do not match
@@ -121,7 +121,7 @@ var (
 		"nvme":    STORAGE_CONTROLLER_NVME,
 		"virtio":  STORAGE_CONTROLLER_VIRTIO,
 		"mmc":     STORAGE_CONTROLLER_MMC,
-		"loop":    STORAGE_CONTROLLER_LOOP,
+		"virtual": STORAGE_CONTROLLER_VIRTUAL,
 	}
 )
 
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -96,7 +96,7 @@ const (
 	STORAGE_CONTROLLER_NVME                      // Non-volatile Memory Express
 	STORAGE_CONTROLLER_VIRTIO                    // Virtualized storage controller/driver
 	STORAGE_CONTROLLER_MMC                       // Multi-media controller (used for mobile phone storage devices)
-	STORAGE_CONTROLLER_VIRTUAL                   // virtual controller, i.e. for a loop device
+	STORAGE_CONTROLLER_LOOP                      // loop device
 )
 
 var (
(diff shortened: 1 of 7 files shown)


Itxaka:
@fromanirh any comments on this? thanks!

jaypipes:
@Itxaka what is the use case for a go-hardware library tracking loopback devices? :)

jaypipes:
> well, we could make the same question about returning mount points and filesystems no? Those are not hardware related at all, more like system status, so thinking in terms of ghw providing info about the current system in which its running, I want to see all the mountpoints, including those from loop devices which are _mounted_. I.E. I want to scan all mountpoints in the system to do something with them, report of scan or whatever, loop devices mounted are part of that system.
> 
> Thats the best I can do to sell this feature LOL :P
> 
> If you want a specific use case, we use loop devices in our installer[0] to test real installs, not really to a real filesystem, but to a loop device so we dont break any real disks, so we do some partition discovery and mounting. We currently use lsblk directly, but it would be much nicer to use a pure golang lib instead of depending on lsblk. Having loop mounts discovery in ghw helps us :)
> 
> [0] https://github.com/rancher-sandbox/elemental

Fair enough. You've sold me on your use case.

I'll do a full review of the PR shortly.

Itxaka:
any update on this @jaypipes ?

pkg/block/block_linux.go
```diff
@@ -329,29 +331,36 @@ func disks(ctx *context.Context, paths *linuxpath.Paths) []*Disk {
...
-			Vendor:                 vendor,
-			Model:                  model,
-			SerialNumber:           serialNo,
-			WWN:                    wwn,
-		}
+		if driveType == DRIVE_TYPE_LOOP && size == 0 {
+			// Do nothing for empty loop devices
+			// There can be several loop devices present on /dev but that doesnt mean they are in use
+			// With this we skip any non-used loop devices
+		} else {
```
jaypipes, line 338: I would prefer to just return all the loop devices that are listed by the OS, regardless of their size. Calling programs can filter out returned results any way they want.

pkg/block/block.go
```diff
@@ -29,6 +29,7 @@ const (
 	DRIVE_TYPE_FDD               // Floppy disk drive
 	DRIVE_TYPE_ODD               // Optical disk drive
 	DRIVE_TYPE_SSD               // Solid-state drive
+	DRIVE_TYPE_LOOP              // loop device
```
jaypipes, line 32: I'm questioning whether "loop" should be a DRIVE_TYPE... is it really a type of drive? I'm not sure... maybe we should make a new `DRIVE_TYPE_VIRTUAL` or `DRIVE_TYPE_PSEUDO`.

pkg/option/option.go
```diff
@@ -137,6 +137,9 @@ type Option struct {
 	// to learn about the system resources.
 	PathOverrides PathOverrides
 
+	// LoopDevices optionally allows reporting disk/partitions that are loop devices
+	LoopDevices *bool
```
jaypipes, line 141: I'm tempted to say let's get rid of this option and just always process loop devices, and let callers filter out any results the want... @fromanirh what do you think?

pkg/block/block_linux.go
Itxaka: Thing is, if you mount something on a loop device, several loop devices will appear in /dev, even if not mounted. This was a kind fo fix to report only those loop devices in use.

```
$ sudo losetup -f --show cOS-Seed-green-0.8.8-g8522da85-x86_64.iso 
/dev/loop0

$ ls /dev/loop*
/dev/loop0  /dev/loop2  /dev/loop4  /dev/loop6  /dev/loop-control
/dev/loop1  /dev/loop3  /dev/loop5  /dev/loop7

$ sudo blkid |grep loop
/dev/loop0: BLOCK_SIZE="2048" UUID="2022-03-23-20-47-00-00" LABEL="COS_LIVE" TYPE="iso9660" PTTYPE="dos"
```

So in reality, in this situation, only the loop0 device is being used, the rest are there for some reason I guess, but not used.  Returning unused loop devices seemed to me a waste of user time filtering those out as they are never gonna need them for anything.

Thoughts?

pkg/block/block.go
jaypipes: OK, let's go with DRIVE_TYPE_VIRTUAL then, @Itxaka.

pkg/block/block_linux.go
jaypipes: @Itxaka OK, I guess I can get behind that reasoning :)

Do me a favor though... in order to reduce the number of lines of code changed here, can you edit the above to this instead?

```go
if storageController == STORAGE_CONTROLLER_LOOP && size == 0 {
    // We don't care about unused loop devices...
    continue
}
```

and please place above the `d := &Disk{...}` definition line.

pkg/option/option.go
jaypipes: @Itxaka yeah, just remove this config option... if there is a huge demand for it, we can always add an option in the future.

fromanirh: Hey! I didn't had time to think this change (overall PR not this specific bit) deep enough. Will do next week. In general I tend to agree to add to the API surface gradually.

Itxaka:
umm, failing on ubuntu but I cant reproduce it locally. As that its picking up the underlaying os devices Im sure there is something wrong in the code..Will try to debug

Itxaka:
Found it, missing a value :D

pkg/block/block.go
```diff
@@ -93,6 +96,7 @@ const (
 	STORAGE_CONTROLLER_NVME                      // Non-volatile Memory Express
 	STORAGE_CONTROLLER_VIRTIO                    // Virtualized storage controller/driver
 	STORAGE_CONTROLLER_MMC                       // Multi-media controller (used for mobile phone storage devices)
+	STORAGE_CONTROLLER_VIRTUAL                   // virtual controller, i.e. for a loop device
```
jaypipes, line 99: Sorry @Itxaka I didn't make myself clear in the original review... I thought `STORAGE_CONTROLLER_LOOP` was good. It's only the `DRIVE_TYPE_LOOP` that I wanted to change to `DRIVE_TYPE_VIRTUAL`... Can we swtich the storage controller back please? :)

jaypipes (changes requested):

pkg/block/block.go
Itxaka: done!

jaypipes (approved):
Very nice work, @Itxaka thank you!

PR #312, opened 2022-03-25
Revert "Detect iSCSI devices"

jaypipes:
This reverts commit 3171315ec3b21f49a6d7f61a02c6a6edb46cf843.

We can discuss whether there are merits to `ghw` determining whether a
SCSI storage controller communicates via a local storage bus or via
TCP/IP (iSCSI), but iSCSI is not a new drive type so I'm reverting this
particular commit.

Signed-off-by: Jay Pipes <jaypipes@gmail.com>


fromanirh (approved):

fromanirh:
looks like we hit a maintenance window: https://www.githubstatus.com/incidents/qt1dj8v3whg9
I think this is safe enough to be merged anyway

avishayt:
Sounds good, thank you

PR #313, opened 2022-04-13
Fixes #267 dont compare pointer addresses

### LINKED ISSUE #267: Could not collect windows disk partition infomation
There is a bug in ghw/pkg/block/block_windows.go 
`https://github.com/jaypipes/ghw/blob/f38192d6ca0e6d937e83bda5fd17f4870b4db014/pkg/block/block_windows.go#L118`
The comparison between `* uint32` and `* uint32` would be wrong in this circumstance.

Bug can be fixed with method below:
```
...
if *diskdrive.Index == *diskpartition.DiskIndex {
    ...
}
```

viocle-kvanek:
Windows partitions will not populate correctly because the disk index on drive and partition are being compared using the pointer address, not the value.

### DIFF
--- a/pkg/block/block_windows.go
+++ b/pkg/block/block_windows.go
@@ -115,7 +115,7 @@ func (i *Info) load() error {
 		}
 		for _, diskpartition := range win32DiskPartitionDescriptions {
 			// Finding disk partition linked to current disk drive
-			if diskdrive.Index == diskpartition.DiskIndex {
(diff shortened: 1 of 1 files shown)


jaypipes:
Hi @viocle-kvanek! Thank you for the contribution! Would you mind signing your commit (`git commit -s --amend`) and re-pushing? The Developer Certificate of Origin (DCO) requires commits be signed.

fromanirh:
the change itself looks good. Let's maybe check (possibly in another PR) that this flow is covered by tests

PR #315, opened 2022-05-01
expose SRIOV information as PCI devices

fromanirh:
Expose informations about SRIOV devices. After some more design iterations (https://github.com/jaypipes/ghw/pull/230#discussion_r755312597), we fixed these goals:

-  model SRIOV devices as special PCI devices, avoiding a new top-level package
-  to shape the API, we start checking what the [k8s SRIOV operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) needs
- to further refine later, we will check what other projects like the [k8s node feature discovery](https://github.com/kubernetes-sigs/node-feature-discovery) consume
- the goal is to enable such projects to switch to ghw if/when they want.

There are few more noteworthy items in this PR:

- due to lack of resources, initial support is for linux only. Nothing linux-specific should have sneaked in the API. Should the current API design make unnecessarily hard to add support on other platforms (e.g. windows) this should be treated as bug
- the code prefers Physical Functions for discovery, but we acknowledge that forcing consumers to traverse the physical functions to learn about virtual functions is awkward, so the API provides shallow references to the Virtual Functions for the sake of practicality.
- this is an alternate implementation of issue #92 , conflicting with #230 

Fixes: https://github.com/jaypipes/ghw/issues/92

Signed-off-by: Francesco Romani [fromani@redhat.com](mailto:fromani@redhat.com)


fromanirh:
this PR is reviewable, but tests are still WIP. The alternate API is pretty much stable though.

fromanirh:
> @fromanirh did you want to rebase this and have it reviewed?

Sure, I'll rebase shortly! other than that this is not very urgent. Up to us if we like more this direction or the original one in #230

fromanirh:
note that most of the tests are missing - deferred until we settle about the direction. Should not be merged without tests, though.

pkg/pci/function.go
```diff
@@ -0,0 +1,109 @@
...
+type Function struct {
+	Device // All Functions are PCI Devices
+	// Parent contains a pointer to the parent physical function.
+	// Will be empty when this is a physical function
+	Parent *Function `json:"parent,omitempty"`
+	// MaxVFs contains the maximum number of supported virtual
+	// functions for this physical function
+	MaxVFs int `json:"max_vfs"`
+	// Functions contains the physical function's virtual functions
+	Functions []*Function `json:"functions"`
```
jaypipes, line 29: Why not just add these attributes to `pkg/pci.Device`?

jaypipes (changes requested):
@fromanirh apologies for the delayed review.

I think I would prefer to have the SRIOV-specific attributes added to the `pkg/pci.Device` struct itself instead of having an "extended" Function struct that derives from Device.

You could have a helper method on `pkg/pci.Info` called `GetSRIOVDevices()` that would return only the PCI devices that were SR-IOV capable. Something like this?

```go
// GetSRIOVDevices returns only the PCI devices that are
// Single Root I/O Virtualization (SR-IOV) capable -- either
// physical of virtual functions.
func (i *Info) GetSRIOVDevices() []*Device {
    res := []*Device{}
    for _, dev := range i.Devices {
        if dev.Parent != nil || len(dev.Functions) != 0 {
            res = append(res, dev)
        }
    }
    return res
}
```

fromanirh:
> @fromanirh apologies for the delayed review.
> 
> I think I would prefer to have the SRIOV-specific attributes added to the `pkg/pci.Device` struct itself instead of having an "extended" Function struct that derives from Device.
> 
> You could have a helper method on `pkg/pci.Info` called `GetSRIOVDevices()` that would return only the PCI devices that were SR-IOV capable. Something like this?
> 
> ```go
> // GetSRIOVDevices returns only the PCI devices that are
> // Single Root I/O Virtualization (SR-IOV) capable -- either
> // physical of virtual functions.
> func (i *Info) GetSRIOVDevices() []*Device {
>     res := []*Device{}
>     for _, dev := range i.Devices {
>         if dev.Parent != nil || len(dev.Functions) != 0 {
>             res = append(res, dev)
>         }
>     }
>     return res
> }
> ```

No worries @jaypipes and thank you for the review. I'm kinda reluctant to conflate even more data in the `pci.Device` struct (if we do than sooner or later we should also dissolve the `gpu` pkg into `pci` - which I'd like! :) ) but I don't feel so strongly to oppose this direction, so let's try and see how it looks.

README.md
```diff
@@ -1123,6 +1123,39 @@ information
...
+	fmt.Printf("%v\n", pci)
+
+	for _, dev := range pci.Functions {
+		fmt.Printf(" %v\n", dev)
+	}
+}
+```
+
+ghw discovers the SRIOV devices scanning PCI devices. Thus, you need to make sure to have scanned the PCI devices before
+to query for SRIOV devices (aka "Functions", "PCI Functions").
```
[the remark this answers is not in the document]
ffromani: fixed, thanks

README.md
```diff
@@ -1123,6 +1123,39 @@ information
...
+Virtual Functions (VFs) are handled by Physical Functions (PFs).
```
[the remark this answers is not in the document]
ffromani: fixed, thanks

README.md
```diff
@@ -1123,6 +1123,39 @@ information
 `ghw.TopologyNode` struct if you'd like to dig deeper into the NUMA/topology
 subsystem
 
+### SRIOV
```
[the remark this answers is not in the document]
ffromani: sure, moved

pkg/pci/function.go
ffromani: on it!

ffromani:
@jaypipes eventually I come to like this approach. We may need a couple more iteration to sort out the details of the API (naming, printing) but overall I'm happy to pursue this direction. PR is updated, rebased and ready for review

ffromani:
depends on https://github.com/jaypipes/ghw/pull/351

README.md
```diff
@@ -879,6 +879,39 @@ Subclass: VGA compatible controller [00]
...
+	pci, err := ghw.PCI()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error getting SRIOV info through PCI: %v", err)
+	}
+
+	fmt.Printf("%v\n", pci)
+
+	for _, dev := range pci.Functions {
+		fmt.Printf(" %v\n", dev)
+	}
```
ffromani, line 905: needs update

ffromani: fixed!

ffromani:
Now that we agreed on the direction, I need to add tests. Will do ASAP.

PR #325, opened 2022-07-16
inaccurate disk size on windows

sspring:
wmi report inaccurate disk size, refer : https://stackoverflow.com/questions/9901792/wmi-win32-diskdrive-to-get-total-sector-on-the-physical-disk-drive  
replace Win32_DiskDrive with DeviceIOControl.

### DIFF
--- a/pkg/block/block_windows.go
+++ b/pkg/block/block_windows.go
@@ -6,7 +6,10 @@
 package block
 
 import (
+	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


fromanirh:
thanks for this contribution @sspring ! could you please add a unit test to ensure this code doesn't regress? None of the active main contributors is a  windows user, so this is especially important, as we rely on CI for correctness.

PR #327, opened 2022-09-19
Fix class name lookups

eliaskoromilas:
All class IDs in https://pci-ids.ucw.cz/v2.2/pci.ids are stored as lowercase hex. The `modalias` file is used to extract the class IDs for a given device. The class IDs in that file can be uppercase HEX, e.g. pci:v00008086d00009D23sv00001028sd000007EAbc**0C**sc05i00. PCI DB lookup fails in that case.

**- What I did**

* Made sure that class-related IDs (class, subclass, programming interface) are parsed as lowercase hex.

**- How I did it**

* Using the `strings.ToLower` function on raw data

### DIFF
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -160,9 +160,9 @@ func parseModaliasData(data string) *deviceModaliasInfo {
 	productID := strings.ToLower(data[18:22])
 	subvendorID := strings.ToLower(data[28:32])
 	subproductID := strings.ToLower(data[38:42])
-	classID := data[44:46]
-	subclassID := data[48:50]
-	progIfaceID := data[51:53]
+	classID := strings.ToLower(data[44:46])
+	subclassID := strings.ToLower(data[48:50])
+	progIfaceID := strings.ToLower(data[51:53])
 	return &deviceModaliasInfo{
 		vendorID:     vendorID,
 		productID:    productID,
--- a/pkg/pci/pci_linux_test.go
+++ b/pkg/pci/pci_linux_test.go
@@ -133,7 +133,10 @@ func TestPCIMarshalJSON(t *testing.T) {
 		t.Fatalf("Expected no error creating PciInfo, but got %v", err)
 	}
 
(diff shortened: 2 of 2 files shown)


fromanirh:
Hi @eliaskoromilas thanks for this contribution. The change looks good to me and the rationale fully makes sense. Do we have a unit test which highlights this behavior, and, if not, could you please add it?
I want to verify one last thing, which is the lowercased identifer don't leak elsewhere in the codebase, so we don't have unintended side effects.

eliaskoromilas:
> Do we have a unit test which highlights this behavior, and, if not, could you please add it?

In your current test data there are no problematic `modalias` files. I could add some real `modalias` file contents (with upper-case class IDs), and verify that are now parsed correctly.

> I want to verify one last thing, which is the lowercased identifer don't leak elsewhere in the codebase, so we don't have unintended side effects.

It affects nothing more than `Class`/`Subclass`/`ProgrammingInterface` struct `ID`s and `Name`s. All of which are purely informational.

eliaskoromilas:
@fromanirh Unit test is included.

fromanirh:
thanks for your updates @eliaskoromilas . I will review as soon as possible.

eliaskoromilas:
> thanks for your updates @eliaskoromilas . I will review as soon as possible.

@fromanirh The MacOS build failure seems to be unrelated. Could we bring this in?

fromanirh (approved):
LGTM

fromanirh:
> > thanks for your updates @eliaskoromilas . I will review as soon as possible.
> 
> @fromanirh The MacOS build failure seems to be unrelated. Could we bring this in?

yes, it seems like a glitch. I'll keep an eye on the lane anyway. Sorry for the the delay

fromanirh:
for the record: the pci.ids database (https://pci-ids.ucw.cz/read/PD/) use lowercase letters. This further confirmation that, when in doubt, converting to lowercase is sensible.

fromanirh:
I'm a bit surprised the official PCI-SIG docs all seems to use the uppercase letters - at least the couple I glanced over. but still.

eliaskoromilas:
We just care about how they are stored in the `pci.ids` file. They are all hex numbers and should be case insensitive.

PR #329, opened 2022-12-13
deps: declare go 1.18 in go.mod

### LINKED ISSUE #328: Consider updating declared Go version in go.mod
The `go.mod` file declares `go 1.15` right now. [Module graph pruning](https://tip.golang.org/ref/mod#graph-pruning) was added in Go 1.17, meaning that other modules can import your module and only pick up the transitive dependencies that are actually reachable based on how they use your code. Right now, importing `ghw` means importing a yaml parser, a unit test framework, and a CLI flag parser because the declared go version predates `go 1.17`.

The tidiest dependency graph will come from using `go 1.18`, as:

> The [go.sum file](https://tip.golang.org/ref/mod#go-sum-files) recorded by [go mod tidy](https://tip.golang.org/ref/mod#go-mod-tidy) for a module by default includes checksums needed by the Go version one below the version specified in its [go directive](https://tip.golang.org/ref/mod#go-mod-file-go). So a go 1.17 module includes checksums needed for the full module graph loaded by Go 1.16, but a go 1.18 module will include only the checksums needed for the pruned module graph loaded by Go 1.17. The -compat flag can be used to override the default version (for example, to prune the go.sum file more aggressively in a go 1.17 module).

Go 1.18 is also the oldest Go release still supported by the Go team, so it's a reasonable choice of language version to declare.

Thanks for all of your hard work!

whereswaldon:
This updates the go.{mod,sum} files to declare a version of Go that supports module graph pruning. The change allows consuming projects to avoid picking up dependencies on modules that ghw only uses for testing or for its CLI.

Fixes #328

Signed-off-by: Chris Waldon <christopher.waldon.dev@gmail.com>

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     - name: set up golang
       uses: actions/setup-go@v2
       with:
-        go-version: 1.17
+        go-version: 1.18
 
     - name: format
       run: ./hack/check-format.sh
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        go: [ '1.15', '1.16', '1.17' ]
+        go: [ '1.18', '1.19']
     steps:
     - uses: actions/checkout@v2
 
@@ -63,7 +63,7 @@ jobs:
     runs-on: windows-2019
     strategy:
       matrix:
-        go: [ '1.16' ]
+        go: [ '1.18' ]
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


fromanirh:
thanks for this PR! I'd be happy to approve considering @jaypipes seems already happy with this.
Could you please check if our github actions configuration needs updates? It should be simple as doing version bumps.
It would be nice to have go.{mod,sum} and CI always aligned

whereswaldon:
I've attempted to update the Go versions for CI.

fromanirh:
Thanks a lot @whereswaldon , looks great! I'm gonna wait a little time to let @jaypipes chime in (e.g. to assess how this change interacts with future releases plans, because we're jumping quite a bit of versions, if at all), if not I'll merge later considering the past conversation in the issue. Thanks again!

jaypipes (approved):
Excellent, thank you so much @whereswaldon! :)

Issue #336, opened 2023-04-21
ARM: failed to determine memory area for node: open /sys/devices/system/memory/block_size_bytes: no such file or directory


wanyaoqi:
`topology.New()` will report:
WARNING: failed to determine memory area for node: open /sys/devices/system/memory/block_size_bytes: no such file or directory

ARM no  /sys/devices/system/memory directory.
As the this comment: https://github.com/kata-containers/runtime/issues/796#issuecomment-425393743
ARM not support memory hotplug，and i can't get physical memory from sysfs.

Is there any way around this problem

jaypipes:
Hi @wanyaoqi thanks for submitting this issue. You can [turn off these warnings](https://github.com/jaypipes/ghw/#disabling-warning-messages). For ARM, topology memory information may not be available due to platform limitations.

wanyaoqi:
Hi @jaypipes , thanks for your reply.  The problem is get numa node physical memory size,  others topo info is works.
And i don't need numa node physical memory size actually, so i just paste usable size to physical size to workaround this problem.  https://github.com/wanyaoqi/ghw/pull/1
Can we check `/sys/devices/system/memory/block_size_bytes` file exists, if not exist return a inaccuracy value( usable size) or return -1 rather than return a error, because other numa topo infomation is worked.

jaypipes:
> Hi @jaypipes , thanks for your reply. The problem is get numa node physical memory size, others topo info is works. And i don't need numa node physical memory size actually, so i just paste usable size to physical size to workaround this problem. [wanyaoqi#1](https://github.com/wanyaoqi/ghw/pull/1) Can we check `/sys/devices/system/memory/block_size_bytes` file exists, if not exist return a inaccuracy value( usable size) or return -1 rather than return a error, because other numa topo infomation is work.

@wanyaoqi we already do that :)

https://github.com/jaypipes/ghw/blob/227643e14380812d61b941f7aa52d1156b6d900a/pkg/memory/memory_linux.go#L113-L119

wanyaoqi:
[...]

PR #337, opened 2023-04-21
assume phys mem == usable mem on ARM

jaypipes:
`/sys/devices/system/memory` does not exist on linux/arm systems that do not support memory hotplug. This means that the
`/sys/devices/system/memory/block_size_bytes` file also does not exist and that is the file that we use to determine the physical RAM associated with a RAM module affined to a NUMA node.

This commit updates the logic in `pkg/memory.AreaForNode` on linux platforms to tolerate the error returned by `memoryBlockSizeBytes()` when `/sys/devices/system/memory/block_size_bytes` does not exist and set a memory `Area`'s `TotalPhysicalBytes` value to the value of `TotalUsableBytes`.

Issue #336

### DIFF
--- a/pkg/memory/memory_linux.go
+++ b/pkg/memory/memory_linux.go
@@ -64,19 +64,31 @@ func AreaForNode(ctx *context.Context, nodeID int) (*Area, error) {
 		fmt.Sprintf("node%d", nodeID),
 	)
 
-	blockSizeBytes, err := memoryBlockSizeBytes(paths.SysDevicesSystemMemory)
+	var err error
+	var blockSizeBytes uint64
+	var totPhys int64
+	var totUsable int64
+
+	totUsable, err = memoryTotalUsableBytesFromPath(filepath.Join(path, "meminfo"))
 	if err != nil {
 		return nil, err
 	}
 
-	totPhys, err := memoryTotalPhysicalBytesFromPath(path, blockSizeBytes)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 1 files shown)


pkg/memory/memory_linux.go
```diff
@@ -64,19 +64,30 @@ func AreaForNode(ctx *context.Context, nodeID int) (*Area, error) {
...
-	totPhys, err := memoryTotalPhysicalBytesFromPath(path, blockSizeBytes)
+	totUsable, err = memoryTotalUsableBytesFromPath(filepath.Join(path, "meminfo"))
 	if err != nil {
 		return nil, err
 	}
 
-	totUsable, err := memoryTotalUsableBytesFromPath(filepath.Join(path, "meminfo"))
-	if err != nil {
-		return nil, err
+	blockSizeBytes, err = memoryBlockSizeBytes(paths.SysDevicesSystemMemory)
```
wanyaoqi, line 77: Hi @jaypipes , I think should use memTotalPhysicalBytes function here, memTotalPhysicalBytes will scan syslog to get physical size, and check return value > 0 to decide whether to use totUsable memory.  
https://github.com/jaypipes/ghw/blob/227643e14380812d61b941f7aa52d1156b6d900a/pkg/memory/memory_linux.go#L105-L113

jaypipes: @wanyaoqi good point. done!

ffromani:
> @ffromani any chance you have some time to review this one? :)

done and merged!

PR #338, opened 2023-04-27
Add AutoNegotiation and PauseFrameUse NICCapabilities

jak3kaj:
Make `AutoNegotiation` and `PauseFrameUse` `NICCapability` objects:
* AutoNegotiation (string value of auto-negotiation)
* PauseFrameUse (string value of pause-frame-use)

Refactoring PR #335

### DIFF
--- a/pkg/net/net_linux.go
+++ b/pkg/net/net_linux.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/jaypipes/ghw/pkg/context"
 	"github.com/jaypipes/ghw/pkg/linuxpath"
+	"github.com/jaypipes/ghw/pkg/util"
 )
 
 const (
@@ -245,14 +246,14 @@ func (nic *NIC) updateNicAttrEthtool(m map[string][]string) {
 	// AutoNegotiation Capability
 	autoNegotiation := NICCapability{Name: "auto-negotiation", IsEnabled: false, CanEnable: false}
 
-	an := strings.Join(m["Auto-negotiation"], "")
-	aan := strings.Join(m["Advertised auto-negotiation"], "")
-	if an && aan {
+	an, anErr := util.ParseBool(strings.Join(m["Auto-negotiation"], ""))
+	aan, aanErr := util.ParseBool(strings.Join(m["Advertised auto-negotiation"], ""))
+	if an && aan && aanErr == nil && anErr == nil {
 		autoNegotiation.IsEnabled = true
 	}
 
-	san := strings.Join(m["Supports auto-negotiation"], "")
-	if san {
+	san, err := util.ParseBool(strings.Join(m["Supports auto-negotiation"], ""))
+	if san && err == nil {
 		autoNegotiation.CanEnable = true
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/net/net_linux.go
```diff
@@ -68,6 +69,8 @@ func nics(ctx *context.Context) []*NIC {
 		nic.MacAddress = mac
 		if etAvailable {
 			nic.Capabilities = netDeviceCapabilities(ctx, filename)
+			// Sets NIC struct fields and appends NICCapability from ethtool output
+			nic.setNicAttrEthtool(ctx, filename)
```
jaypipes, line 73: Can we place the call to `setNicAttrEthtool` within the `netDeviceCapabilities` function?

pkg/net/net_linux.go
```diff
@@ -220,3 +223,115 @@ func netDevicePCIAddress(netDevDir, netDevName string) *string {
...
+	if err != nil {
+		msg := fmt.Sprintf("could not grab NIC link info for %s: %s", dev, err)
+		ctx.Warn(msg)
+		return err
+	}
+
+	m := parseNicAttrEthtool(&out)
+	nic.updateNicAttrEthtool(m)
+
+	return nil
```
jaypipes, line 242: I think it would be better to just merge this into the `netDeviceCapabilities()` function. Then you can re-use the `out` and `path` variables from that function instead of calling again `exec.LookPath` and creating another `bytes.Buffer`

jaypipes (changes requested):
Thanks @jak3kaj! Few suggestions inline for you but this is on a good track!

jak3kaj:
By moving `setNicAttrEthtool()` into `netDeviceCapabilities()` I should have resolved the concerns you shared.

I made some more changes as the logic made less sense after moving the code.  I split up the `updateNicAttrEthtool()` method and created `autoNegCap()` and `pauseFrameUseCap()` so they behaved more like the other code in `netDeviceCapabilities()` (returning `*NICCapabilties` objects).

I put the logic to parse `ethtool` before parsing `ethtool -k` so there is a sort of logical order (ASCII-betical).

If I were to replace running `ethtool` with an `ethtool-like` [go library,](https://pkg.go.dev/github.com/safchain/ethtool) hopefully this function (and functions it calls) are all that would need to be modified.  In the back of my mind I imagine that calling a go library can return the same `map` that `parseNicAttrEthtool()` is returning, so we can use `autoNegCap()` and `pauseFrameUseCap()` in the same way.

jaypipes (approved):
Great work on this @jak3kaj thank you! :) Welcome to the ghw contributor club!

jaypipes:
@jak3kaj can you run `gofmt -w .` on your source tree checkout please? the formatting check is complaining...

jak3kaj:
The thing it's flagging is not flagged by `gofmt` or `go fmt`.  

It wants this:
```go
            expected: []*NICCapability{
                {
                    Name:      "auto-negotiation",
                    IsEnabled: true,
                    CanEnable: true,
                },
                {
                    Name:      "pause-frame-use",
                    IsEnabled: false,
                    CanEnable: false,
                },
            },

```

instead of this:
```go
            expected: []*NICCapability{
                &NICCapability{
                    Name:      "auto-negotiation",
                    IsEnabled: true,
                    CanEnable: true,
                },
                &NICCapability{
                    Name:      "pause-frame-use",
                    IsEnabled: false,
                    CanEnable: false,
                },
            },
```

🤷 I will bow to the formatter script

jak3kaj:
oops, I just fixed the commit that wasn't signed off and rebased with main.  This should be ready to merge now.

PR #340, opened 2023-05-02
Fix #339 by getting the physical disk info which properly shows SSD vs HDD

### LINKED ISSUE #339: HDD is classified as SSD on windows 10
I was trying out the lib on windows 10 and was surprised to see that it was misclassifying my HDD as an SSD.  Turns out the currently logic just says all "fixed" media types are SSD, which is incorrect.  The "fixed" media type simply means the disk is internal to the PC, so its not "removable" media for example like a USB drive.  I have what I believe to be the correct code (working for me locally) that I will create a PR for.

williambrode:
Fixes #339

### DIFF
--- a/pkg/block/block_windows.go
+++ b/pkg/block/block_windows.go
@@ -6,12 +6,36 @@
 package block
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/StackExchange/wmi"
 
 	"github.com/jaypipes/ghw/pkg/util"
 )
+
+type physicalDiskMediaType int
+
+const (
+	PHYSICAL_DISK_MEDIA_TYPE_UNSPECIFIED physicalDiskMediaType = 0
+	PHYSICAL_DISK_MEDIA_TYPE_HDD         physicalDiskMediaType = 3
+	PHYSICAL_DISK_MEDIA_TYPE_SSD         physicalDiskMediaType = 4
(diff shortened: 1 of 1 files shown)


pkg/block/block.go
```diff
@@ -32,6 +32,29 @@ const (
...
+		return DRIVE_TYPE_UNKNOWN
+	case PHYSICAL_DISK_TYPE_HDD:
+		return DRIVE_TYPE_HDD
+	case PHYSICAL_DISK_TYPE_SSD:
+		return DRIVE_TYPE_SSD
+	case PHYSICAL_DISK_TYPE_SCM:
+		return DRIVE_TYPE_UNKNOWN
+	}
+	return DRIVE_TYPE_UNKNOWN
+}
```
jaypipes, line 56: can we move this into `pkg/block/block_windows.go` please, since it's Windows-specific? Also, let's name this `physicalDiskMediaType` (unexported) to denote that it's the WMI physical disk media type code.

williambrode: Ah yeah, I had it there before and then moved it for some reason... will move it back.

jaypipes (approved):
Awesome, thank you @williambrode! :)

jaypipes:
@williambrode I think a rebase might be needed here.

jaypipes (approved):

Issue #341, opened 2023-05-03
Linux:  failed to determine memory area for node: open /sys/devices/system/node/node0/memory_failure/state: no such file or directory


matildeY:
Failed to get topologyNodes return error
`failed to determine memory area for node: open /sys/devices/system/node/node0/memory_failure/state: no such file or directory`
 
 1- since latest linux kernel there is a new directory called "memory_failure" that matches pattern defined in function `memoryTotalPhysicalBytesFromPath`  and makes it failed.
 
 2- is it possible to call "memTotalPhysicalBytes" instead of "memoryTotalPhysicalBytesFromPath" in function "AreaForNode"  which contains a fallback function to the syslog approach ?

jaypipes:
> hi, thanks for your answer, before submitting this bug i've tried with the main branch and the error is still present.
> 
> The memoryTotalUsableBytesFromPath call is ok and does not return error

OK, thank you for verifying that @matildeY! I will push a fix ASAP.

asiffer:
Hi! I got the same warning when calling `ghw.GPU` :
`WARNING: failed to determine memory area for node: open /sys/devices/system/node/node0/memory_failure/state: no such file or directory`
(it is hidden when I use the `ghw.WithDisableWarnings()` option)

jaypipes:
> Hi! I got the same warning when calling `ghw.GPU` : `WARNING: failed to determine memory area for node: open /sys/devices/system/node/node0/memory_failure/state: no such file or directory` (it is hidden when I use the `ghw.WithDisableWarnings()` option)

Hi @asiffer! What version of `ghw` are you using? I believe I fixed that issue but have not yet cut a new release of `ghw` that includes it.

asiffer:
`v0.10.0` in my `go.mod` file

jaypipes:
@asiffer I just cut a new release: https://github.com/jaypipes/ghw/releases/tag/v0.11.0

Wait a bit for Go mod/packaging to publish the new package and then update your go.mod. Should be good to go after that! :)

asiffer:
Awesome 🤩  Thanks a lot!

ffromani:
@asiffer @matildeY 
Hi! could you please kindly confirm version v0.11.0 fixes your issues, so we can close this bug? Thanks!

asiffer:
Yes the last version fixed it 👍🏻 Thanks again

matildeY:
yes, it's ok for me. Thanks

[...]

PR #343, opened 2023-05-06
only process memory cell/block subdirs in sysfs

jaypipes:
The `/sys/devices/system/memory` and `/sys/devices/system/node/nodeX` subdirectories contains one or more subdirectories that begin with the word "memory" and end in a 0-based cell/block index for that memory. e.g. `/sys/devices/system/node/node0/memory63` is a directory containing information files about the 64th memory block in NUMA node 0.

Previously, code that gathered total physical memory by looking at these subdirectories was using a simple glob on `memory*` to determine those memory block subdirectories. However, in some recent Linux kernels a `/sys/devices/system/memory/memory_failure` file causes that simple glob to backfire. This patch replaces the simple glob with a read of the `/sys/devices/system/memory` or `/sys/devices/system/node/nodeX` directory and regex matches on `memory\d$` to determine if the subdirectory is a memory block one.

Fixes Issue #341

### DIFF
--- a/pkg/memory/memory_linux.go
+++ b/pkg/memory/memory_linux.go
@@ -38,6 +38,10 @@ var (
 	// System log lines will look similar to the following:
 	// ... kernel: [0.000000] Memory: 24633272K/25155024K ...
 	_REGEX_SYSLOG_MEMLINE = regexp.MustCompile(`Memory:\s+\d+K\/(\d+)K`)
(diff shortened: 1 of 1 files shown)


ffromani:
LGTM

waiting for the missing CI lane (macos-1025) to complete/rerun before to merge

ffromani (approved):
LGTM

(still!)

PR #344, opened 2023-05-08
update our GH actions runners/workflows

jaypipes:
Brings in Ubuntu 22.04, Windows Server 2022 and MacOS 12 (instead of the now-deprecated <=MacOS-11 runner).

Also updates the Go versions to test Go 1.18-1.20 on various platforms.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   # tier 0: system-independent checks
   format:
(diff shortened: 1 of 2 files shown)


jaypipes:
note to @ffromani... I ran `gofmt -w .` to regenerate the doc.go file since the fmt linter check in Go 1.19+ has a different output.

ffromani (approved):
LGTM

Issue #345, opened 2023-06-27
Possible bug in parsing core.ID and core.Index


cyga:
In changes between v0.10.0 to v0.11.0
(more precisely in `git diff 59f01e7caa962f36aef7293abaab6a4f770391ba 34f45f4cfdb628931eb1732fa414a7ae97191354`)
the following changes were introduced:

```
v0.10.0:
    ID = "core id" from /proc/cpuinfo
    Index = index in /proc/cpuinfo
v0.11.0: 
    ID = not set (0)
    Index = value from /sys/devices/system/cpu/cpuXX/topology/core_id
```

Is it a bug or a feature?

1) Documentation wasn't changed for it:

```
ghw.ProcessorCore.ID is the uint32 identifier that the host gave this core. Note that this does not necessarily equate to a zero-based index of the core within a physical package. For example, the core IDs for an Intel Core i7 are 0, 1, 2, 8, 9, and 10
ghw.ProcessorCore.Index is the zero-based index of the core on the physical processor package
```

2) If this is designed behaviour can you at least keep ID same as Index (as far as I can see, "core id" has the same value on my linuxes as "/sys/devices/system/cpu/cpuXX/topology/core_id") for some period. So, we can adopt this change during transition period?

jaypipes:
Hi @cyga! Would you mind providing some details about your host system? Specifically what architecture are you using? To answer your question, no, it was not the intention to change these field values, though the implementation of CPU information gathering was indeed [refactored in the v0.11.0 timeframe](https://github.com/jaypipes/ghw/commit/34f45f4cfdb628931eb1732fa414a7ae97191354). /cc @rockrush @glimchb

glimchb:
from a quick look without any testing yet looks like this might be a fix:
```
diff --git a/pkg/cpu/cpu_linux.go b/pkg/cpu/cpu_linux.go
index d4b048d..274ca05 100644
--- a/pkg/cpu/cpu_linux.go
+++ b/pkg/cpu/cpu_linux.go
@@ -138,7 +138,7 @@ func processorsGet(ctx *context.Context) []*Processor {
                coreID := util.SafeIntFromFile(ctx, coreIdPath)
                core := CoreByID(proc.Cores, coreID)
                if core == nil {
-                       core = &ProcessorCore{Index: coreID, NumThreads: 1}
+                       core = &ProcessorCore{ID: coreID, Index: len(proc.Cores), NumThreads: 1}
[...]

PR #346, opened 2023-07-04
linux: refactor ProcessorCore retrieval

jaypipes:
Reworks the way that ProcessorCore structs are determined on Linux by breaking up the long `processorsGet` function into smaller functions and using standard maps for looking up Processor and logicalProcessor structs.

Removes the `ProcessorCore.Index` field entirely, since it was confusing and did not actually represent anything useful. Added clarity to the documentation about `ProcessorCore.LogicalProcessors` and how those numbers are actually the zero-based index of processors on the host and are sometimes called "thread siblings" and that zero-based index of logical processor IDs has nothing to do with the core ID.

Fixes Issue #345

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,12 +408,13 @@ A `ghw.ProcessorCore` has the following fields:
   core. Note that this does *not* necessarily equate to a zero-based index of
   the core within a physical package. For example, the core IDs for an Intel Core
   i7 are 0, 1, 2, 8, 9, and 10
-* `ghw.ProcessorCore.Index` is the zero-based index of the core on the physical
-  processor package
 * `ghw.ProcessorCore.NumThreads` is the number of hardware threads associated
   with the core
-* `ghw.ProcessorCore.LogicalProcessors` is an array of logical processor IDs
-  assigned to any processing unit for the core
+* `ghw.ProcessorCore.LogicalProcessors` is an array of ints representing the
+  logical processor IDs assigned to any processing unit for the core. These are
+  sometimes called the "thread siblings". Logical processor IDs are the
+  *zero-based* index of the processor on the host and are *not* related to the
+  core ID.
 
 ```go
 package main
--- a/pkg/block/block_test.go
+++ b/pkg/block/block_test.go
@@ -53,9 +53,6 @@ func TestBlock(t *testing.T) {
 	if d0.Name == "" {
 		t.Fatalf("Expected disk name, but got \"\"")
 	}
-	if d0.SerialNumber == "unknown" {
-		t.Fatalf("Got unknown serial number.")
-	}
 	if d0.SizeBytes <= 0 {
 		t.Fatalf("Expected >0 disk size, but got %d", d0.SizeBytes)
 	}
--- a/pkg/cpu/cpu_linux.go
+++ b/pkg/cpu/cpu_linux.go
@@ -37,117 +37,111 @@ func (i *Info) load() error {
 	return nil
 }
 
-func ProcByID(procs []*Processor, id int) *Processor {
-	for pid := range procs {
-		if procs[pid].ID == id {
-			return procs[pid]
-		}
-	}
-	return nil
-}
-
-func CoreByID(cores []*ProcessorCore, id int) *ProcessorCore {
-	for cid := range cores {
-		if cores[cid].Index == id {
-			return cores[cid]
-		}
-	}
-	return nil
-}
-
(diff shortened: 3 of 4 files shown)


ffromani:
ack @jaypipes I'll be happy to have a look ASAP

jaypipes:
@ffromani anything about this look faulty to you? I'd like to cut a new release with this fix in the next few days.

README.md
```diff
@@ -408,12 +408,13 @@ A `ghw.ProcessorCore` has the following fields:
   core. Note that this does *not* necessarily equate to a zero-based index of
   the core within a physical package. For example, the core IDs for an Intel Core
   i7 are 0, 1, 2, 8, 9, and 10
-* `ghw.ProcessorCore.Index` is the zero-based index of the core on the physical
-  processor package
```
ffromani, line 412: what's our process for removing fields from the API?
Do we need any special care (deprecation) ?

pkg/block/block_test.go
```diff
@@ -53,9 +53,6 @@ func TestBlock(t *testing.T) {
 	if d0.Name == "" {
 		t.Fatalf("Expected disk name, but got \"\"")
 	}
-	if d0.SerialNumber == "unknown" {
-		t.Fatalf("Got unknown serial number.")
-	}
```
ffromani, line 58: related?

pkg/cpu/cpu_linux.go
```diff
@@ -37,117 +37,109 @@ func (i *Info) load() error {
...
 
-	// Iterate on /sys/devices/system/cpu/cpuN, not on /proc/cpuinfo
-	Entries, err := ioutil.ReadDir(paths.SysDevicesSystemCPU)
+	// /sys/devices/system/cpu pseudodir contains N number of pseudodirs with
+	// information about the logical processors on the host. These logical
+	// processor pseudodirs are of the pattern /sys/devices/system/cpu/cpu{N}
+	fnames, err := ioutil.ReadDir(paths.SysDevicesSystemCPU)
 	if err != nil {
-		return nil
+		ctx.Warn("failed to read /sys/devices/system/cpu: %s", err)
```
ffromani, line 52: ok, the rest of the function should be skipped as `fnames` is expected to be empty slice

pkg/cpu/cpu_linux.go
```diff
@@ -37,117 +37,109 @@ func (i *Info) load() error {
...
+// processor ID
+func coreIDFromLogicalProcessorID(ctx *context.Context, lpID int) int {
+	paths := linuxpath.New(ctx)
+	// Fetch CPU ID
+	path := filepath.Join(
+		paths.SysDevicesSystemCPU,
+		fmt.Sprintf("cpu%d", lpID),
+		"topology", "core_id",
+	)
+	return util.SafeIntFromFile(ctx, path)
```
ffromani, line 142: unrelated to this PR: I think eventually we should evaluate if better to cache data from sysfs instead of hitting the FS layer all the time. Granted, sysfs is a pseudofs, but still

ffromani (commented):
LGTM but with the question regarding the API change
Other comments are minor and not-blocking.

Is the test coverage for the linux CPU parsing enough to cover the changes? could be a good chance to add/expand tests. Can be done in another PR.
@jaypipes feel free to merge pending the decision if/when add tests.

README.md
jaypipes: In the past I've left poorly-named fields (like `Id` instead of `ID`) in the structs in case people were using the old names and then removed them after a deprecation period. However, in this case, the old `Index` field really was confusing and wrong: it stated that it was the zero-based index of the core on the physical processor package, but that was incorrect. It was *one of* the zero-based indexes of a *logical processor* associated with the physical core... and so I figure we should just remove this field entirely and force users to change their code if they were referencing `ProcessorCore.Index`. Thoughts?

pkg/block/block_test.go
jaypipes: This was just a bad unit test that I removed. Sorry, unrelated to this PR.

pkg/cpu/cpu_linux.go
jaypipes: ah, good catch. I will re-add the `return nil`

ffromani:
@jaypipes feel free to merge if you are happy with what we are converging on https://github.com/jaypipes/ghw/pull/346#discussion_r1258228853

PR #347, opened 2023-07-14
marshal: replace `github.com/ghodss/yaml` with `gopkg.in/yaml.v2`

Juneezee:
Hi :wave: 

The `github.com/ghodss/yaml` package is no longer actively maintained. There are numerous inquiries about the project's status on its issue tracker: https://github.com/ghodss/yaml/issues.

Since we are only using the `JSONToYAML` from `github.com/ghodss/yaml`, we can implement the code ourselves [^1] and remove an extra dependency.

---

Alternatively, we can replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`, which is a permanent fork currently maintained by Kubernetes SIG. Please let me know if you prefer using a package for YAML unmarshalling, and I will gladly update this pull request.

[^1]: https://github.com/ghodss/yaml/blob/0ca9ea5df5451ffdf184b4428c902747c2c11cd7/yaml.go#L45-L61

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/jaypipes/pcidb v1.0.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v0.0.3
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.1
 	howett.net/plist v1.0.0
 )
 
--- a/go.mod
+++ b/go.mod
@@ -4,10 +4,10 @@ go 1.18
 
 require (
 	github.com/StackExchange/wmi v1.2.1
-	github.com/ghodss/yaml v1.0.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -4,10 +4,10 @@ go 1.18
 
 require (
 	github.com/StackExchange/wmi v1.2.1
-	github.com/ghodss/yaml v1.0.0
 	github.com/jaypipes/pcidb v1.0.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v0.0.3
+	gopkg.in/yaml.v2 v2.4.0
```
jaypipes, line 10: How about `gopkg.in/yaml.v3` instead? :)

Juneezee: Updated to `gopkg.in/yaml.v3`

jaypipes (approved):
Awesome, thank you @Juneezee! :)

PR #350, opened 2023-07-21
clean CI jobs, Makefile and go.mod/Dockerfile

jaypipes:
harden and refactor GH action for unit tests
    
    Updates the GH Action workflow for unit testing with security best
    practices, including reduce permissions, the step security action
    hardener, and using SHA-specific Action releases.

separate linter job in Github Actions
    
    pulls out the golangci-lint job from the main go.yml Github Action and
    into its own workflow that follows security best practices like ensuring
    read-only permissions and using SHA-specific Action releases.
    
cleanup Makefile, Dockerfile, go.mod
    
    Updates to Go 1.19 in go.mod to get us to a more modern Go release.
    Similarly changes the Dockerfile that builds the `ghwc` binary to use Go
    1.19 and use a hashed version of the Alpine base image to address
    security best practices.
    
    Finally, removes the outdated and unnecessary Make targets and
    unnecessarily complicated calls to find Go packages. These targets and
    fancy code was left over from before Go modules and when we were using
    `dep` and vendoring things.
    
    Adds a `.github/workflows/fmtcheck.yml` that follows GH actions security
    best practices and separates the fmtcheck stuff from the other tests in
    the main `.github.com/workflows/go.yml`.
    
remove old hack/check-format.sh and doc.go content
    
    The hack/check-format.sh script is not used anywhere (and there is a
    `make fmtcheck` target that accomplishes the same thing). So, removing
    that.
    
    Also remove all the duplicative (and out-of-date) documentation from the
    `doc.go` file for two reasons:
    
    1) pkg.go.dev now includes the README in its entirety and therefore
       there is not a need to duplicate the information from the README into
       a `doc.go` file.
    2) Having two places for examples and docs just meant the information in
       `doc.go` quickly got out of date.


jaypipes:
@ffromani this PR does a bunch of cleanup of our CI jobs (GH Action runners) in preparation for the v1 milestone. Most of the cleanup was around hardening of the workflows in accordance with security best practices from step-security. Feel free to review! :)

.github/workflows/fmtcheck.yml
```diff
@@ -0,0 +1,30 @@
...
+  fmtcheck:
+    runs-on: ubuntu-latest
+    steps:
+     - name: harden runner
+       uses: step-security/harden-runner@55d479fb1c5bcad5a4f9099a5d9f37c8857b2845 # v2.4.1
+       with:
+         egress-policy: block
+         disable-sudo: true
+         allowed-endpoints: >
+           github.com:443
```
ffromani, line 22: nice!

ffromani (approved):
very nice work @jaypipes !

PR #353, opened 2023-08-29
Change udev properties for Linux disk serial

avishayt:
The properties that are currently used are not always accurate, for example we have seen that they sometimes contain the WWN and not the serial. This change uses SCSI_IDENT_SERIAL and ID_SCSI_SERIAL first. This behavior aligns with what lsblk does:
https://github.com/util-linux/util-linux/blob/36c52fd14b83e6f7eff9a565c426a1e21812403b/misc-utils/lsblk-properties.c#L122-L128

### DIFF
--- a/pkg/block/block_linux.go
+++ b/pkg/block/block_linux.go
@@ -147,6 +147,16 @@ func diskSerialNumber(paths *linuxpath.Paths, disk string) string {
 		return util.UNKNOWN
 	}
 
+	// First try to use the serial from sg3_utils
+	if serial, ok := info["SCSI_IDENT_SERIAL"]; ok {
+		return serial
+	}
+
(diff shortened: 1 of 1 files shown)


jaypipes:
Hi @avishayt! Thanks very much for this contribution! Would you mind signing the commit in order for the DCO check to succeed? I'll be reviewing the PR shortly.

avishayt:
Thanks!  Done.

jaypipes:
looks like there's some transient GH actions runner issues. Gonna wait a little while and retrigger. should be fine after.

jaypipes:
@avishayt the CI failures are due to StepSecurity GH Actions checkers not allowing egress to proxy.golang.org. Nothing to do with your PR. Thanks much for your contribution! :)

Issue #357, opened 2023-12-30
Support querying current resource usage


jaypipes:
`ghw` was originally designed as a system capabilities library, allowing developers to query for system functionality, NUMA topology and hardware capabilities. We deliberately did *not* query for resource usage, instead calling out in the documentation that there [were other systems that were designed for monitoring](https://github.com/jaypipes/ghw/blob/835f6a233b6db0d384a7e5a254eef4f05cdcae5e/README.md#inspecting--monitoring).

However, as the years have gone by, I've noticed that many systems that import `ghw` (and a few that call the `ghwc` example CLI tool) also import other libraries or shell out to many other tools (depending on the host system) to additionally collect resource usage information. And, I think it would be used for these consumer tools to be able to use `ghw` for *both* discovering capabilities as well as collecting current resource usage amounts. That way, these consumer tools don't need to shell out to other tools and can use a single library for collecting resource capability and usage information.

ffromani:
I'm in favor of this enhancement. I wonder if we can implement the new functionality as add-on libraries on top of existing ghw packages, leaving the latter mostly unchanged. Perhaps is just more practical to bake in the new functionalities in the existing codebase?

PR #358, opened 2023-12-30
begin work of supporting current resource usage

jaypipes:
This patch begins the process of supporting current resource usage information in `ghw`. The first resource we support is memory on Linux systems.

Users can use the `ghw.WithCollectUsage(true)` function to instruct `ghw` to gather usage information for the module in question. Alternately, users can use the `GHW_COLLECT_USAGE=1` environs variable. The following shows the effect of the variable:

```
jaypipes@lappie:~/src/github.com/jaypipes/ghw$ go run cmd/ghwc/main.go memory
memory (16GB physical, 16GB usable)
jaypipes@lappie:~/src/github.com/jaypipes/ghw$ GHW_COLLECT_USAGE=1 go run cmd/ghwc/main.go memory
memory (16GB physical, 16GB usable, 9GB used)
```

Issue #357


ffromani (commented):
At glance, looks fine. The one and only reservation I have is that I'm not a fan of optional fields which are filled only if a global flag is set. I think we could have a couple or more of these fields in the various subsystems.
 I'd consider an approach on which we add current resource usage in a more isolated way with respect the current return value. A very quick proposal could be to wrap the resource usage in substructs.
I'll have a proper review shortly in few days time tops.

jaypipes:
> At glance, looks fine. The one and only reservation I have is that I'm not a fan of optional fields which are filled only if a global flag is set. I think we could have a couple or more of these fields in the various subsystems. I'd consider an approach on which we add current resource usage in a more isolated way with respect the current return value. A very quick proposal could be to wrap the resource usage in substructs. I'll have a proper review shortly in few days time tops.

@ffromani Hi! :)

Yeah, I had a similar thought and actually originally I had an approach where I had a separate `Usage` struct for each module (to go along with the `Info` structs in the modules. But eventually I just thought it would be more straightforward to add the fields to the main `Info` structs themselves.

For things like memory usage, we will always need to determine the structure (NUMA topology) of the system itself and then add usage per NUMA node (on Linux via `/sys/devices/system/node/nodeX/meminfo`) so we are always going to need to grab the `Info` structs... so I thought it easier to just add usage fields to the `Info` structs themselves... that way, we don't need to do any alternate JSON/YAML serialization or all the plumbing for the `Info.load()` methods...

jaypipes:
@ffromani we could also consider collecting usage information by default and making the collection of usage information disable-able via flag?

Issue #361, opened 2024-02-21
Retrieval of core_id and physical_package_id for CPUs that are offline. 


kishen-v:
It was observed that, on CPUs that are made offline using `echo 0 > /sys/devices/system/cpu/cpuX/online`, the `topology` directory does not contain information related to core_id and physical_package_id. 

Due to the missing file, a related message `WARNING: failed to read int from file: open /sys/devices/system/cpu/cpuX/topology/physical_package_id: no such file or directory` is raised, and a similar behaviour is also observed if the `core_id` file is missing too.

In the case of a large compute node, this may lead to plenty of warnings printed along with other information if many CPUs are offline. 

One such instance which was observed on the system is
```
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu10/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu10/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu11/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu11/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu12/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu12/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu13/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu13/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu14/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu14/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu15/topology/physical_package_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu15/topology/core_id: no such file or directory
WARNING: failed to read int from file: open /sys/devices/system/cpu/cpu16/topology/physical_package_id: no such file or directory

... - From cpu8 to cpu64.
```

This can be overcome by checking the `online` file in the `cpuX` directory for the bit `1` before proceeding to access the physical_package_id/core_id file.

PR #362, opened 2024-02-21
Read core and physical package ID of the CPUs that are online

kishen-v:
Fixes: https://github.com/jaypipes/ghw/issues/361

Added check to fetch the physical_package_id/core_id after determining if the CPU is online.


ffromani:
LGTM

any chance to add a test or two to exercise this code?

kishen-v:
Hi @ffromani, I've added a couple of tests to check if the warning is being raised in the case of missing files (core_id/physical_package_id) in the topology directory.  Could you PTAL when time permits?

Thanks!

pkg/cpu/cpu_linux.go
```diff
@@ -64,6 +65,10 @@ func processorsGet(ctx *context.Context) []*Processor {
 			continue
 		}
 
+		onlineFilePath := filepath.Join(paths.SysDevicesSystemCPU, fmt.Sprintf("cpu%d", lpID), onlineFile)
+		if util.SafeIntFromFile(ctx, onlineFilePath) == 0 {
```
jaypipes, line 69: this will return -1 if the file does not exist... perhaps it's worth changing this to:

```go
if util.SafeIntFromFile(ctx, onlineFilePath) != 1 {
```

in order to account for when there are problems locating the `/online` file?

pkg/cpu/cpu_linux.go
```diff
@@ -201,6 +206,10 @@ func CoresForNode(ctx *context.Context, nodeID int) ([]*ProcessorCore, error) {
 			)
 			continue
 		}
+		onlineFilePath := filepath.Join(cpuPath, onlineFile)
+		if util.SafeIntFromFile(ctx, onlineFilePath) == 0 {
```
jaypipes, line 210: ditto for here.

jaypipes (commented):
Hi @kishen-v thanks very much for your contribution! One little suggestion inline for you...

pkg/cpu/cpu_linux.go
kishen-v: I agree, will make the required changes.

Thanks!

kishen-v: It was observed that this change didn't fit well for the ARM architecture, as the `online` file does not exist under the  cpuX directory. This lead to the mismatch in the expected vs observed tests results causing a breaking change. 

The `online` file exists for the amd64/ppc64le architecture where, the corresponding files are in the topology directory may/may not exist based on the bit set in the `online` file.  I have made changes to account only if the cpu is offline, before the files under the topology directory are checked. This doesn't seem to affect the `arm64` architecture as the `-1` returned from the caller doesn't affect the rest of the code execution. 

Thanks!

jaypipes (approved):

PR #369, opened 2024-06-01
uplift Go, dependencies and action runners

jaypipes:
Uplifts our Go version to 1.21, our pcidb and cobra dependencies to the latest version and uplifts our Github action runners appropriately.


go.mod
```diff
@@ -1,22 +1,22 @@
 module github.com/jaypipes/ghw
 
-go 1.19
+go 1.21
```
ffromani, line 3: do we need any feature of go >= 1.20, or do one of our uplifted deps?

.github/workflows/fmtcheck.yml
```diff
@@ -22,6 +22,7 @@ jobs:
            github.com:443
            api.github.com:443
            proxy.github.com:443
+           proxy.golang.org:443
```
ffromani, line 25: why is this needed? I can't recall when the proxy was introduced but I seem to recall it was prior to 1.20

ffromani (commented):
I'm about to merge but I have a couple hopefully quick questions as inline comments

go.mod
jaypipes: @ffromani no, not particularly. I was just keeping us up to date with more modern Go :)

.github/workflows/fmtcheck.yml
ffromani: I have no major objections here so I guess we can move on

go.mod
ffromani: I'm a fan of keeping backward compatibility as much as possible for library code, but I don't have any real objection here so let's bump it.
Perhaps next time let's try to be a tad more conservative for the sake of making life of the consumers of ghw easier

Issue #371, opened 2024-07-29
Add HugePages info to AreaForNode


Traumeel:
I am using Topology functionality to get info about cpu and mem in NUMA architecture. What I am currently missing is data about HugePages. At least HugePages_Total, HugePages_Free and currently set size.

Maybe you are willing to extend logic here

https://github.com/jaypipes/ghw/blob/main/pkg/memory/memory_linux.go#L63

and add info about HugePages.

In case I am missing something and this info can be obtained by some other means in the lib, please feel free to redirect me.

Cheers!

jaypipes:
@ffromani @Traumeel apologies for not responding earlier! I must have missed this in my inbox :(

Yes, this seems like a straightforward enhancement. @Traumeel would you like to submit this feature or would you like one of us to handle it?

ffromani:
No probs @jaypipes I noticed late as well :) 
Just in case, I'd be happy to help as backup.

Traumeel:
> @ffromani @Traumeel apologies for not responding earlier! I must have missed this in my inbox :(

@jaypipes No worries, all good)

> Yes, this seems like a straightforward enhancement. @Traumeel would you like to submit this feature or would you like one of us to handle it?

I did PR, please have a look.

jaypipes:
Closing this one out. The next release of ghw will include your feature, @Traumeel, thank you! :)

PR #374, opened 2024-09-15
add HugePage info to memory and topology

Traumeel:
https://github.com/jaypipes/ghw/issues/371 fix

Add huge page info to Topology and Memory functionality.

### DIFF
--- a/pkg/memory/memory.go
+++ b/pkg/memory/memory.go
@@ -29,8 +29,8 @@ type Module struct {
 	Vendor       string `json:"vendor"`
 }
 
-// HugePage describes huge page info
-type HugePage struct {
+// HugePageAmounts describes huge page info
+type HugePageAmounts struct {
 	Total   int64 `json:"total"`
 	Free    int64 `json:"free"`
 	Surplus int64 `json:"surplus"`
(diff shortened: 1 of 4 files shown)


pkg/memory/memory.go
```diff
@@ -37,8 +46,14 @@ type Area struct {
 	TotalPhysicalBytes int64 `json:"total_physical_bytes"`
 	TotalUsableBytes   int64 `json:"total_usable_bytes"`
 	// An array of sizes, in bytes, of memory pages supported in this area
-	SupportedPageSizes []uint64  `json:"supported_page_sizes"`
-	Modules            []*Module `json:"modules"`
+	SupportedPageSizes []uint64 `json:"supported_page_sizes"`
+	// Default system huge pages size, in bytes
+	DefaultHugePageSize uint64 `json:"default_huge_page_size"`
+	// Amount of memory, in bytes, consumed by huge pages of all sizes.
+	HugeTLBSize int64 `json:"huge_tlb_size"`
```
jaypipes, line 53: the TLB is the lookaside buffer used for looking up the huge page physical location from a virtual address. It's also a Linux-specific thing and I'd prefer to keep the Area struct system-agnostic as much as possible.

How about we call this field `TotalHugePageBytes`?

pkg/memory/memory.go
```diff
@@ -37,8 +46,14 @@ type Area struct {
 	TotalPhysicalBytes int64 `json:"total_physical_bytes"`
 	TotalUsableBytes   int64 `json:"total_usable_bytes"`
 	// An array of sizes, in bytes, of memory pages supported in this area
-	SupportedPageSizes []uint64  `json:"supported_page_sizes"`
-	Modules            []*Module `json:"modules"`
+	SupportedPageSizes []uint64 `json:"supported_page_sizes"`
+	// Default system huge pages size, in bytes
```
jaypipes, line 50:
```suggestion
	// Default system huge page size, in bytes
```

pkg/memory/memory.go
```diff
@@ -29,6 +29,15 @@ type Module struct {
 	Vendor       string `json:"vendor"`
 }
 
+// HugePage describes huge page info
+type HugePage struct {
```
jaypipes, line 33: Let's call this struct `HugePageAmounts` since it's not describing the huge page itself but rather a set of distinct finite amounts of huge pages of a particular size within a memory area.

jaypipes (changes requested):
@Traumeel thank you so much for the PR! :) I left a few notes inline for you but overall, I like this quite a bit!

Traumeel:
> @Traumeel thank you so much for the PR! :) I left a few notes inline for you but overall, I like this quite a bit!

@jaypipes thank you for the swift review) I am glad you like it!

pkg/memory/memory.go
Traumeel: fixed

pkg/memory/memory.go
Traumeel: fixed

pkg/memory/memory.go
Traumeel: fixed

jaypipes (approved):
thanks @Traumeel !

Issue #375, opened 2024-09-20
emulated graphics cards or framebuffers cause GPU test failure on Linux


jaypipes:
There's a few things going on with the GPU test when running on my new system76 monster workstation:

```
➜  ghw git:(fix-gpu-test) go test -v ./pkg/gpu/...
=== RUN   TestGPUWithoutNUMANodeInfo
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu0/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu1/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu10/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu11/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu2/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu3/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu4/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu5/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu6/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu7/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu8/online: no such file or directory
WARNING: failed to read int from file: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/cpu9/online: no such file or directory
WARNING: failed to determine memory area for node: open /tmp/ghw-gpu-testing-298741871/sys/devices/system/node/node0/meminfo: no such file or directory
[...]

PR #376, opened 2024-09-20
supply override PCIDB_PATH env var in unit tests

jaypipes:
Some of the unit tests inspect snapshot filesystems and for the GPU Info-collection, the `github.com/jaypipes/pcidb` library is used to get PCI information. That library no longer (as of version v0.9) does an automatic fetch of the canonical pci.ids file if it cannot find a pci.ids file in the local filesystem. It no longer does that network fetch because the upstream maintainer of the pci.ids database noticed too many fetches coming from projects that imported `ghw` which transitively imported `pcidb`.

Because of this change in behaviour of no longer doing an automatic network fetch of the pci.ids database, a unit test in `ghw` that was relying on the pcidb library was no longer working. This patch fixes two things related to the PCI handling in the `pkg/gpu` package. First, it cleans up the parsing of the `/sys/class/drm` symlink contents by looking through the path parts and attempting a regex match for a PCI address. Second, it adds an env-var override to the GPU unit tests to point `github.com/jaypipes/pcidb` to a specific pci.ids file that I've included in the `testdata/` directory.

Closes Issue #375


jaypipes:
@ffromani this one fixes up a nagging failure in unit tests that I found..

ffromani (approved):
I think we can improve further later on.
e.g. I don't like too much os.Stat + SafeIntFromFile - I'll try to improve.
But indeed, later on: this work LGTM and fixes the issue, so I'm fine merging.

jaypipes:
> I think we can improve further later on. e.g. I don't like too much os.Stat + SafeIntFromFile - I'll try to improve. 

totally agreed :) I was actually going in to do some cleanup in that area when I ran into this. I'll be pushing up some cleanup PRs later this weekend!

jaypipes:
I've got to fix up the GH action workflows since they are stuck... will do that in a separate PR...

PR #377, opened 2024-09-21
update Step Security hardener action

jaypipes:
The step security Github Action step was stuck and not letting the
workflow proceed. Step Security advised updating to the latest
harden-runner version.
    
Issue step-security/harden-runner#466


jaypipes:
@ffromani note I [submitted an issue to the step-security/harden-runner project](https://github.com/step-security/harden-runner/issues/466) to see if they can advise us what's going on with their GH action step.

Issue #380, opened 2024-09-23
publish JSONSchema for our JSON/YAML output


jaypipes:
In preparation for a v1.0 release, I'd like to publish a versioned JSONSchema document that consumers may validate against (and also allow us to evolve our API over time). Part of our public API is the JSON/YAML that we produce when people marshal the ghw structs into JSON or YAML, so I'm conscious of changes to that public API. Having a versioned JSONSchema will allow us to deprecate and evolve things over time in a more controlled fashion.

jaypipes:
@ffromani thoughts on this one? :)

ffromani:
I like a lot and we need to add roundtrip tests!

PR #381, opened 2024-09-23
use TotalXXX instead of NumXXX

jaypipes:
All other fields that contain a total amount of something begin with the word `Total` except for the `pkg/cpu.Processor.NumCores` and `NumThreads` fields. This commit adds new fields to `pkg/cpu.Processor` called `TotalCores` and `TotalHardwareThreads` to bring the names of these fields in-line with all other amount/total fields.

In addition, adds a `TotalHardwareThreads` field to the `pkg/cpu.Info` struct and deprecates the less-explicit `TotalThreads` field.

I'd like to remove the now-deprecated `NumThreads`, `NumCores`, and `TotalThreads` fields before we get to a `v.1.0` series.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ information about the CPUs on the host system.
 
 * `ghw.CPUInfo.TotalCores` has the total number of physical cores the host
   system contains
-* `ghw.CPUInfo.TotalThreads` has the total number of hardware threads the
-  host system contains
+* `ghw.CPUInfo.TotalHardwareThreads` has the total number of hardware threads
+  the host system contains
 * `ghw.CPUInfo.Processors` is an array of `ghw.Processor` structs, one for each
   physical processor package contained in the host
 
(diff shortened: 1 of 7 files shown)


ffromani (commented):
LGTM.
- we do change the docs, but that's fine, we're recommending the new fields
- we keep setting correctly the old fields for backward compat

we do change the JSON representation though, do we need to test/guarantee the JSON roundtrip?
@jaypipes feel free to merge anytime if the JSON representation is not a concern for backward compat

jaypipes:
> LGTM.
> 
>     * we do change the docs, but that's fine, we're recommending the new fields
> 
>     * we keep setting correctly the old fields for backward compat
> 
> 
> we do change the JSON representation though, do we need to test/guarantee the JSON roundtrip? @jaypipes feel free to merge anytime if the JSON representation is not a concern for backward compat

@ffromani yeah, see [the issue I created yesterday](https://github.com/jaypipes/ghw/issues/380) about the JSON/YAML representation and firming that up with a published JSONSchema. :)

jaypipes:
@ffromani also, note that I only *added* fields to the JSON output in this PR. It's still backwards-compat representation of the JSON/YAML but I do want to deprecate things before v1.0

PR #387, opened 2024-10-24
Populate `Vendor` field for ARM cpus

zeeke:
In ARM  systems, the `/proc/cpuinfo` file does not contain `vendor_id` fields. In the `lscpu` tool, the `CPU implementor` code and a lookup table are used to provide a human-readable vendor name [1].

When running on ARM systems, populate the `CPU.Info.Processors[*].Vendor` field with the "ARM" string.

[1] https://github.com/util-linux/util-linux/pull/564


jaypipes (approved):
Rock on! Thanks very much for the patch, @zeeke ! :)

Issue #394, opened 2024-12-04
Add filtering functionality for processing accelerators


mlorenzofr:
Continuing the work done in #383, in some scenarios it would be interesting to have the option to filter the processing accelerator hardware by vendor, model or PCI class.

ffromani:
thanks! will review ASAP

PR #395, opened 2024-12-04
Add custom filters for processing accelerators

mlorenzofr:
This PR implements the feature to add a filter for hardware acceleration detection as described in issue #394.
This is just a draft, let's review it and we can change anything we feel is appropriate :wink:


README.md
```diff
@@ -966,6 +966,16 @@ Each `ghw.AcceleratorDevice` struct contains the following fields:
...
+#### filters
+The `ghw.Accelerator()` function accepts a slice of filters, of type string, as parameter
+in format `[<vendor>]:[<device>][:<class>]`, (same is the _lspci_ command).
+
+Some filter examples:
+* `::0302`. Select 3D controller cards.
+* `10de::0302`. Select Nvidia (`10de`) 3D controller cards (`0302`).
+* `1da3:1060:1200`. Select Habana Labs (`1da3`) Gaudi3 (`1060`) processing accelerator cards (`1200`).
+* `1002::`. Select AMD ATI hardware.
+
```
ffromani, line 978: the filtering functionality is something we want, no questions about it.
I imagined it a bit higher level though. Like, pass a single function which takes a `pci.Device` object and return `(bool, string)` to convey if it should be accepted, and if not why. Or perhaps we can just (ab)use a `error` as return value: if `nil`, accept, if not `err.Error()` can tell us why.

ffromani: something like
```
func Accelerator(filters ...func(dev *pci.Device) bool)
```

pkg/accelerator/accelerator_linux.go
```diff
@@ -60,13 +62,69 @@ func (i *Info) load() error {
 		if !isAccelerator(device) {
 			continue
 		}
-		accelDev := &AcceleratorDevice{
-			Address:   device.Address,
-			PCIDevice: device,
+		if len(i.DiscoveryFilters) > 0 {
```
ffromani, line 65: the way I envisioned is that the user should be able to feed more filters in addition to the unavoidable buliltin
```isAccelerator := func(dev *pci.Device) bool {}```

probably using a different interface to maximize the safety: user-provided filters should never be allowed to modify the `pci.Devices` passed to them (or, equivalently, any change made should be discarded)

ffromani (commented):
initial review

/cc @jaypipes

PR #397, opened 2025-01-28
ci: golang: rotate the most 3 recent supported versions

ffromani:
we want to support most 3 recent golang versions
on modern platform and the most modern golang version available on legacy platforms, so let's bump the
versions accordingly


jaypipes (approved):
thanks @ffromani ! much appreciated :)

ffromani:
> thanks @ffromani ! much appreciated :)

happy to help! the end goal is to start consuming ARM runners on github CI, to unblock ARM support which is a very requested item (see: https://github.com/jaypipes/ghw/pull/396 very much WIP)

PR #398, opened 2025-01-28
gh action: update macos lane

ffromani:
per https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories the oldest support  macos version is nowadays 13,
so use it instead of no longer available v12.

In addition, bump the golang version on macos to a more respectable 1.21 instead of ancient 1.18
(see also: https://github.com/jaypipes/ghw/pull/397)


jaypipes (approved):
:+1: thanks @ffromani !

PR #399, opened 2025-02-02
fix: Replace strings.TrimPrefix with filepath.Rel

Gofastasf:
- filepath.Rel to find relative path in platform aware way.
- file closing by defer.


jaypipes (approved):
Thanks very much for your PR, @Gofastasf ! :) Looks good to me!

PR #400, opened 2025-02-02
fix: Refactor OpenDestination func

Gofastasf:
- Replaced os.Stat call and os.Create with os.OpenFile with O_EXCL to avoid TOCTOU.
- Improved logic to handle file size check only when the file exists.


jaypipes (approved):
another nice cleanup, thank you @Gofastasf :)

PR #401, opened 2025-02-02
Improve string and file extension handling: ReplaceAll & filepath.Ext

Gofastasf:
Replaced strings.Replace with updated strings.ReplaceAll

Use filepath.Ext for file extension instead of strings.HasSuffix.


jaypipes (approved):
Sorry for the long delay in merging this @Gofastasf, approved.

PR #406, opened 2025-03-03
deps: accelerator: remove deps

ffromani:
golang 1.21.z already features `slices.Contains`
(see: https://pkg.go.dev/slices@go1.21.13#Contains) so we can use the stdlib instead of adding a depenency just for a single function.
No intended change in behavior.


ffromani:
right, we still support golang 1.18. I 'd write a trivial `Contains` replacement with a comment to switch to `slice.Contains` once the minimum supported version is 1.21.
I feel pulling 2 deps for a such trivial code is too much in our case.

jaypipes (approved):
@ffromani works for me! :) thanks!

PR #409, opened 2025-03-12
Get WWN for multipath block devices

zaneb:
Device Mapper devices have `DM_WWN` set in udev in place of `ID_WWN_WITH_VENDOR_EXTENSION`. Reading `DM_WWN` allows us to see the WWN for multipath fibre channel or iSCSI block devices.

Because `DM_WWN` contains the vendor extension, there is no corresponding change to `WWNNoExtension`.


jaypipes (approved):
LGTM, thanks @zaneb :)

PR #410, opened 2025-03-18
Add more attributes that match CPU model on Linux

infastin:
Following attributes in `/proc/cpuinfo` can be used to get CPU model name on some of the architectures:
- Processor  — ARM
- cpu model  — MIPS and ARM
- Model Name — LoongArch


jaypipes (approved):
Hey thank you @infastin for the contribution! :)

PR #412, opened 2025-04-22
Set partition UUID from VolumeSerialNumber

baier233:
> **VolumeSerialNumber**  
> Data type: **string**  
> Access type: **Read-only**  
> 
> The volume serial number of the logical disk.  
> Example: "A8C3-D032"
> 
> ― [Win32_LogicalDisk docs](https://learn.microsoft.com/en-us/windows/win32/cimwin32prov/win32-logicaldisk)


pkg/block/block_windows.go
```diff
@@ -173,13 +174,14 @@ func (i *Info) load() error {
...
 							p := &Partition{
+								Disk:       disk,
 								Name:       strings.TrimSpace(*logicaldisk.Caption),
 								Label:      strings.TrimSpace(*logicaldisk.Caption),
 								SizeBytes:  *logicaldisk.Size,
 								MountPoint: *logicaldisk.DeviceID,
 								Type:       *diskpartition.Type,
 								IsReadOnly: toReadOnly(*diskpartition.Access),
-								UUID:       "",
+								UUID:       *logicaldisk.VolumeSerialNumber,
```
jaypipes, line 184: @baier233 I think the `UUID` field should only contain UUIDs, so I'm not sure that VolumeSerialNumber is appropriate for that field of Partition. What about either adding a new SerialNumber field to the Partition struct *or* using the [existing `FilesystemLabel` field](https://github.com/jaypipes/ghw/blob/b2e06de36a815f788443cde2b8c6e8688f2434ac/pkg/block/block.go#L271-L273) instead?

baier233: I fully agree that Volume Serial Numbers are not RFC-compliant UUIDs.

I noticed that the Linux implementation of `ghw` uses `udev` to fetch the `ID_PART_ENTRY_UUID` attribute for partitions. This means that for non-GPT partitions, the UUID field will not be RFC-compliant (see [systemd PR #3293](https://github.com/systemd/systemd/pull/3293)).

As documented in [Debian’s Part-UUID](https://wiki.debian.org/Part-UUID):
> "FAT and NTFS filesystems have shorter names (e.g., sda6 and sdb1), but they are still listed by UUID."

To maintain backward compatibility with existing applications that expect partition identifiers in the `UUID` field, I believe it would be appropriate to use `VolumeSerialNumber` in this field.

baier233: > @baier233 OK, makes sense to me :) Could you add a note to the [Partition.UUID field's docstring](https://github.com/jaypipes/ghw/blob/b2e06de36a815f788443cde2b8c6e8688f2434ac/pkg/block/block.go#L268-L270) that mentions this may not be a UUID format for Windows partitions?

Sure,I've also updated the README description for `ghw.Partition.UUID` to reflect the Windows support:

jaypipes (approved):
awesome, thank you @baier233 ! :)

PR #424, opened 2025-08-29
Add USB devices

christoph-zededa:
This PR adds USB devices to the list of devices.

The code basically iterates over `/sys/bus/usb/devices`, follows the symlinks and reads `uevent`, `interface` and `product`.

For the snapshot, the symlinks in `/sys/bus/usb/devices` are added as well as the aforementioned files.


christoph-zededa:
Fixed the build errors for Mac OS X and the linting errors about missing error-return-checks.

PR #430, opened 2025-10-27
pci: add iommu group

christoph-zededa:
the iommu group is important when using pci-passthrough as the whole group has to be passed through (unless using some hacks). Therefore it is an important piece of information that should be part of `pci.Device`.


jaypipes:
@christoph-zededa not entirely sure what happened to this PR... did you push commits from a different branch into yours? Typically what you want to do is `git rebase` to pull in commits from a main or upstream branch.

PR #431, opened 2025-10-28
pci: add parent device address

christoph-zededa:
the parent device address is important, too (like IOMMU group), when using pci-passthrough as the both devices should be passed through together


ffromani (commented):
I'm fine adding the parent address. Quick review, I'm probably overly cautious over backward compatibility

jaypipes (approved):
all looks good to me, thank you @christoph-zededa :)

europaul (commented):
LGTM

PR #432, opened 2025-11-06
Add WithDisableTopology option to reduce memory consumption

halfcrazy:
This commit adds a new WithDisableTopology() option that allows users to skip system topology detection when calling ghw.PCI() or ghw.GPU(), significantly reducing memory consumption for use cases that don't require NUMA topology information.

Performance impact on real production servers:
> Dual-socket server (2 NUMA nodes, 96 cores):
BenchmarkPCIMemoryComparison/WithTopologyDetection 100	 100708875 ns/op	29332384 B/op	  289365 allocs/op
BenchmarkPCIMemoryComparison/WithDisableTopology 100	  58473485 ns/op	17201283 B/op	  247006 allocs/op
> - Memory: 41% reduction (28 MB -> 16.4 MB)
> - Time: 42% faster (100.7 ms -> 58.5 ms)
> - Allocations: 15% fewer (289,365 -> 247,006)
> 
> Quad-socket server (4 NUMA nodes, 128 cores):
BenchmarkPCIMemoryComparison/WithTopologyDetection 100	 448132406 ns/op	344566098 B/op	  257009 allocs/op
BenchmarkPCIMemoryComparison/WithDisableTopology 100	  64284435 ns/op	18707107 B/op	  193877 allocs/op
> - Memory: 94.6% reduction (328 MB -> 17.8 MB)
> - Time: 85.7% faster (448 ms -> 64 ms)
> - Allocations: 24.6% fewer (257,009 -> 193,877)
> - Saves 310 MB per call - critical for large deployments
>

### DIFF
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -137,22 +137,32 @@ func (i *Info) String() string {
 // New returns a pointer to an Info struct that contains information about the
 // PCI devices on the host system
 func New(opt ...option.Option) (*Info, error) {
-	topo, err := topology.New(opt...)
-	if err != nil {
-		return nil, fmt.Errorf(
-			"failed to initialize PCI info dur to failure to initialize "+
-				"Topology info: %w",
-			err,
-		)
-	}
 	opts := &option.Options{}
 	for _, o := range opt {
 		o(opts)
 	}
+
 	// by default we don't report NUMA information;
 	// we will only if are sure we are running on NUMA architecture
 	info := &Info{
-		arch: topo.Architecture,
+		arch: topology.ArchitectureSMP, // default to SMP
+	}
+	// Skip topology detection if requested to reduce memory consumption
+	if opts.DisableTopology {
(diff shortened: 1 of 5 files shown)


pkg/pci/pci.go
```diff
@@ -137,22 +137,32 @@ func (i *Info) String() string {
...
+		if err != nil {
+			return nil, fmt.Errorf(
+				"failed to initialize PCI info due to failure to initialize "+
+					"Topology info: %w",
+				err,
+			)
+		}
+		info.arch = topo.Architecture
+	}
+	if opts.PCIDB != nil {
```
halfcrazy: I found `options.PCIDB` no longer used, i'm not sure if this is expected.

jaypipes: > Author

@halfcrazy yes, we've totally changed the way that options and context are handled. I will rework this PR soon, I promise. sorry for the delay.

PR #463, opened 2026-05-12
pci: expose more sysfs metadata on Device for tree walks

dims:
Adds the per-device information callers need to walk the PCI tree and inspect each node beyond vendor / product / class identifiers:

  * Modalias        - raw modalias string from sysfs
  * Parent/Children - resolved *Device pointers, built after
                      enumeration; ParentAddress remains the
                      serialized form
  * VPD()           - lazy reader for /sys/.../vpd with a TLV parser
                      handling the 0x82 / 0x90 / 0x91 / 0x78 large
                      resource tags and PN / EC / SN / V0-VZ
                      keywords
  * Subsystems()    - per-device map of subsystem kind to
                      registered device names for
                      drm / infiniband / net / nvme
  * SetVPD()        - test seam to pre-populate the VPD cache

Tests cover the VPD parser via in-memory byte fixtures, and exercise Modalias / Parent pointer / NVMe Subsystems against the existing linux-amd64-intel-i7-1270P snapshot.

### DIFF
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -9,6 +9,7 @@ package pci
 import (
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	"github.com/jaypipes/pcidb"
 
@@ -41,6 +42,25 @@ type Device struct {
 	// for IOMMU Groups see also:
 	// https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-iommu-deep-dive
 	IOMMUGroup string `json:"iommu_group"`
+	// Modalias is the raw contents of the device's sysfs `modalias` file.
+	// Useful for callers that want to do custom vendor/device/class matching
+	// beyond what the pcidb lookup provides.
+	Modalias string `json:"modalias,omitempty"`
+
+	// Parent is the resolved parent Device pointer for this device (the
+	// PCIe upstream port or root complex device). It is nil for root
+	// devices. Populated after enumeration; not included in JSON output to
+	// avoid cycles. Use ParentAddress for the serialized form.
+	Parent *Device `json:"-"`
+	// Children are the resolved downstream Device pointers, in
+	// no particular order. Populated after enumeration; not included in
+	// JSON output.
+	Children []*Device `json:"-"`
+
+	// Cached VPD result (lazy). Guarded by vpdOnce.
+	vpdOnce sync.Once
+	vpd     *VPD
+	vpdErr  error
 }
 
 type devIdent struct {
@@ -60,6 +80,7 @@ type devMarshallable struct {
 	Subclass      devIdent `json:"subclass"`
 	Interface     devIdent `json:"programming_interface"`
 	IOMMUGroup    string   `json:"iommu_group"`
+	Modalias      string   `json:"modalias,omitempty"`
 }
 
 // NOTE(jaypipes) Device has a custom JSON marshaller because we don't want
@@ -97,6 +118,7 @@ func (d *Device) MarshalJSON() ([]byte, error) {
 			Name: d.ProgrammingInterface.Name,
 		},
 		IOMMUGroup: d.IOMMUGroup,
+		Modalias:   d.Modalias,
 	}
 	return json.Marshal(dm)
 }
--- a/pkg/pci/pci_linux.go
+++ b/pkg/pci/pci_linux.go
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	// found running `wc` against real linux systems
-	modAliasExpectedLength = 54
+	// modalias content length, sans trailing newline
+	modAliasExpectedLength = 53
 )
 
 func (i *Info) load(ctx context.Context) error {
@@ -94,6 +94,15 @@ func getDeviceIommuGroup(paths *linuxpath.Paths, pciAddr *pciaddr.Address) strin
 	return filepath.Base(dest)
 }
 
+func getDeviceModalias(paths *linuxpath.Paths, pciAddr *pciaddr.Address) string {
+	fp := getDeviceModaliasPath(paths, pciAddr)
+	data, err := os.ReadFile(fp)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSpace(string(data))
+}
+
 func getDeviceParentAddress(paths *linuxpath.Paths, pciAddr *pciaddr.Address) string {
 	devPath := filepath.Join(paths.SysBusPciDevices, pciAddr.String())
 
@@ -133,18 +142,6 @@ type deviceModaliasInfo struct {
 	classID      string
 	subclassID   string
 	progIfaceID  string
-}
-
-func parseModaliasFile(fp string) *deviceModaliasInfo {
-	if _, err := os.Stat(fp); err != nil {
-		return nil
-	}
-	data, err := os.ReadFile(fp)
-	if err != nil {
-		return nil
-	}
-
-	return parseModaliasData(string(data))
 }
 
 func parseModaliasData(data string) *deviceModaliasInfo {
@@ -393,14 +390,13 @@ func (info *Info) getDevices(ctx context.Context) []*Device {
 			return nil
 		}
 
-		// no cached data, let's get the information from system.
-		fp := getDeviceModaliasPath(paths, pciAddr)
-		if fp == "" {
+		modalias := getDeviceModalias(paths, pciAddr)
+		if modalias == "" {
 			log.Warn(ctx, "error finding modalias info for device %q", address)
 			return nil
 		}
 
-		modaliasInfo := parseModaliasFile(fp)
+		modaliasInfo := parseModaliasData(modalias)
 		if modaliasInfo == nil {
 			log.Warn(ctx, "error parsing modalias info for device %q", address)
 			return nil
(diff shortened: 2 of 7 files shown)


jaypipes:
Well hello there @dims! :) thanks very much for the contribution, sir!

dims:
> Well hello there @dims! :) thanks very much for the contribution, sir!

Hello my dear friend :)

pkg/pci/vpd_linux.go
```diff
@@ -0,0 +1,206 @@
...
+//	FG  Fabric Geography
+//	LC  Location
+//	MN  Manufacturer ID
+//	PG  Extended Capability
+//	SN  Serial Number
+//	V0-VZ  Vendor-specific
+//	Y0-YZ  System-specific
+//
+// Vendor-specific keywords are returned in ReadOnly/ReadWrite keyed by
+// the two-character keyword name (e.g. "V3").
```
jaypipes, line 47: Appreciate the good docstring!

pkg/pci/pci_linux.go
```diff
@@ -94,6 +94,19 @@ func getDeviceIommuGroup(paths *linuxpath.Paths, pciAddr *pciaddr.Address) strin
...
+	data, err := os.ReadFile(fp)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSpace(string(data))
+}
+
+func getDeviceSysdir(paths *linuxpath.Paths, pciAddr *pciaddr.Address) string {
+	return filepath.Join(paths.SysBusPciDevices, pciAddr.String())
+}
```
jaypipes: I don't think we need a helper function for this. Probably best to just use the more explicit `filepath.Join()` call, like is done on line 111.

pkg/pci/pci.go
```diff
@@ -41,6 +42,31 @@ type Device struct {
...
+	// Children are the resolved downstream Device pointers, in
+	// no particular order. Populated after enumeration; not included in
+	// JSON output.
+	Children []*Device `json:"-"`
+
+	// sysdir is the absolute path to the device's sysfs directory (e.g.
+	// "/sys/bus/pci/devices/0000:01:00.0"). Set during enumeration when the
+	// devices are discovered via sysfs. Empty for devices constructed via
+	// ParseDevice. Lazy accessors (VPD, Subsystems) use this.
+	sysdir string
```
jaypipes: Generally, I've avoided using private member fields like this in the exported structs exposed by `ghw`. Since this is really only just caching the result of a single call to `filepath.Join()`, I think I'd rather omit this.

pkg/pci/pci.go
```diff
@@ -41,6 +42,31 @@ type Device struct {
...
 	IOMMUGroup string `json:"iommu_group"`
+	// Modalias is the raw contents of the device's sysfs `modalias` file.
+	// Useful for callers that want to do custom vendor/device/class matching
+	// beyond what the pcidb lookup provides.
+	Modalias string `json:"modalias,omitempty"`
+
+	// Parent is the resolved parent Device pointer for this device (the
+	// PCIe upstream port or root complex device). It is nil for root
+	// devices. Populated after enumeration; not included in JSON output to
+	// avoid cycles — use ParentAddress for the serialized form.
```
jaypipes: Do I see an emdash in the above code comment? Hello, Claude! ;)

pkg/pci/pci_linux.go
jaypipes: Maybe replace the above code with a call to the `getModaliasFile()` function you added above?

pkg/pci/subsystems.go
```diff
@@ -0,0 +1,66 @@
...
+// directories that PCI devices may expose under their sysfs entry.
+// Callers should treat this list as advisory — devices may expose
+// other kinds (e.g. "scsi_host", "block") that this package does not
+// enumerate by default.
+var SubsystemKinds = []string{
+	"drm",
+	"infiniband",
+	"net",
+	"nvme",
+}
```
jaypipes: See comment below..  let's rename this to `linuxSystemKinds`. Also, please either rename this file to `pkg/pci/subsystems_linux.go` or put this in `pkg/pci/pci_linux.go` since it's Linux-specific.

pkg/pci/subsystems.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Subsystems returns a map keyed by subsystem kind (one of
```
jaypipes: I've got a few concerns:

1. Use of the term `subsystem` here is confusing considering there is an orthogonal concept called a PCI Subsystem that has to do with child (or sub-) devices that may be provided by a different PCI vendor (product). Indeed, the `pcidb` library (imported from `ghw`) exposes the [`Subsystems()` method on the `pcidb.Product` struct](https://github.com/jaypipes/pcidb/blob/3134fab14c12d5dea455ea08c1b1843c8eaebc06/types/product.go#L21).
2. Since this is only for Linux, the appropriate thing to do would be to add a stub method implementation in `pci_stub.go` that returns nothing.
3. This method returns "device names" but PCI has no such identifying concept. That's really just a Linux-specific human-readable identifier/link to the PCI device address. So, perhaps it's better to not have this be a method on `pkg/pci.Device` at all and instead make a function, exposed ONLY in `pci_linux.go` called `DeviceNamesByLinuxSystem()` like so:

```go
// DeviceNamesByLinuxSystem returns a map, keyed by well-known Linux system string ("drm",
// "infiniband", "net", "nvme") of lists of Linux device names associated with the supplied PCI
// Device.
func DeviceNamesByLinuxSystem(ctx context.Context, dev *Device) map[string][]string {
    paths := linuxpath.New(ctx)
    address := dev.Address
    if address == "" {
        return nil
    }
    pciAddr := pciaddr.FromString(address)
    devPath := filepath.Join(paths.SysBusPciDevices, pciAddr.String())
    for _, kind := range linuxSystemKinds {
		entries, err := os.ReadDir(filepath.Join(devPath, kind))
		if err != nil {
			continue
		}
		names := make([]string, 0, len(entries))
		for _, e := range entries {
			names = append(names, e.Name())
		}
		if len(names) == 0 {
			continue
		}
		sort.Strings(names)
		out[kind] = names
	}
	return out
}
```

pkg/pci/vpd.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return out, nil
+}
+
+func trimTrailingNuls(s string) string {
+	for len(s) > 0 && (s[len(s)-1] == 0 || s[len(s)-1] == ' ') {
+		s = s[:len(s)-1]
+	}
+	return s
+}
```
jaypipes: Heh, thanks Claude, but the idiomatic and most efficient way to do this is:

```go
strings.TrimSpace(s)
```

pkg/pci/vpd.go
```diff
@@ -0,0 +1,206 @@
...
+// SetVPD pre-populates the VPD cache for the device. Subsequent calls
+// to VPD() return the supplied value (and nil error) without touching
+// sysfs. Intended for tests and for callers that construct Device
+// instances by hand.
+func (d *Device) SetVPD(v *VPD) {
+	d.vpdOnce.Do(func() {
+		d.vpd = v
+		d.vpdErr = nil
+	})
+}
```
jaypipes: I don't think this method is used. Best to remove it, since we don't use setters like this in other ghw structs and the `vpd` and `vpdErr` fields are populated on first execution of the `VPD()` method.

pkg/pci/vpd_linux.go
```diff
@@ -0,0 +1,206 @@
+//
```
jaypipes, line 4: Since this is specific to Linux, please rename this file to `vpd_linux.go`

jaypipes (changes requested):
Thanks @dims for the great submission! Left a few comments inline for you to have a look at :)

pkg/pci/vpd_linux.go
dims: Done in ccc0662 — full rename to vpd_linux.go with //go:build linux.

pkg/pci/vpd.go
dims: Removed. Let's get it the way you want and then in a follow up i'll share what i am trying to do and we can discuss options

pkg/pci/vpd.go
dims: 🙃

pkg/pci/subsystems.go
dims: All three addressed.

dims: 1. The method is gone
2. pci_stub.go returns nil on non-Linux.
3. The function name avoids "Subsystem" entirely

pkg/pci/subsystems.go
dims: subsystems_linux.go with a //go:build linux tag, and the list is unexported as linuxSystemKinds.

pkg/pci/pci_linux.go
dims: Done!

pkg/pci/pci.go
dims: :)

pkg/pci/pci.go
dims: done

pkg/pci/pci_linux.go
dims: Done

dims:
Done @jaypipes !

jaypipes (approved):
Awesomesauce, thank you so much @dims! :)

pkg/pci/pci_linux.go
```diff
@@ -94,6 +94,15 @@ func getDeviceIommuGroup(paths *linuxpath.Paths, pciAddr *pciaddr.Address) strin
 	return filepath.Base(dest)
 }
 
+func getDeviceModalias(paths *linuxpath.Paths, pciAddr *pciaddr.Address) string {
+	fp := getDeviceModaliasPath(paths, pciAddr)
+	data, err := os.ReadFile(fp)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSpace(string(data))
```
jaypipes, line 103: @dims the only thing I can think of as to why the unit tests might be failing here is that we're trimming space from both leading and trailing instead of just trailing... Maybe switch this to `strings.TrimSuffix(string(data), " ")` and see if that fixes the errors popping up in the unit tests?

dims: yep. fixing!

dims:
@jaypipes tested in a couple of boxes. please trigger workflows again

dims:
thanks @jaypipes !

FYI, was prototyping something https://github.com/dims/rdma_topo

context: https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/issues/1123#issuecomment-4430055309

jaypipes:
> FYI, was prototyping something https://github.com/dims/rdma_topo

I think that might be a private repo? :)

dims:
> I think that might be a private repo? :)

public now!

dims:
@jaypipes if you see something that can belong to this repo from there, please let me know and i'll open another PR

PR #464, opened 2026-05-14
pci: make Device.VPD respect the caller's chroot

dims:
Device.VPD() reads /sys/bus/pci/devices/<addr>/vpd via
linuxpath.New(context.Background()), so it ignores any chroot the
parent Info was constructed with. Take ctx as a parameter and use it
for the linuxpath lookup; callers can now pass a context built with
option.WithChroot and the VPD read will respect it.
context.Background() preserves the previous /sys behavior.

### DIFF
--- a/pkg/pci/pci.go
+++ b/pkg/pci/pci.go
@@ -9,7 +9,6 @@ package pci
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


jaypipes (approved):
Yup, much better :)

PR #465, opened 2026-05-14
pci: add Walk, Ancestors and Root helpers on Device

dims:
PR #463 exposed Parent and Children pointers on pci.Device so callers
can walk the tree. Anything non-trivial consuming those fields ends
up reinventing the same three traversals.


dims:
feels like this will be more useful outside of my repo.

jaypipes (approved):
I agree this is generally useful and I think it's fine to add these methods to the `pkg/pci.Device` struct.

PR #466, opened 2026-05-14
use ContextFromEnv() as the base context when args are passed

jimmykarily:
Commit 4e7ac7a restored pre-0.23.0 behaviour for the zero-args call path, but the with-args path still starts from context.TODO() and silently drops environment variables like GHW_CHROOT, GHW_DISABLE_WARNINGS, and GHW_DISABLE_TOOLS.

In pre-0.23.0 ghw, option.Merge always called EnvOrDefaultChroot() (and its siblings) as defaults regardless of which options the caller passed, so e.g. `block.New(WithDisableTools())` still picked up GHW_CHROOT. The 0.23.0 refactor lost that, breaking downstream callers that set the env var and expected it to compose with other options. A common pattern is test mocks that export GHW_CHROOT and rely on consumers (including indirect ones via libraries that call block.New with a couple of modifiers) to honour it.

Use ContextFromEnv() as the base inside ContextFromArgs so env-derived values survive unless overridden by an explicit Modifier, Option, or context.Context arg.

Includes regression tests covering each of: Modifier-only args, WithChroot override of env, old-style option.WithChroot override of env, and explicit context.Context replacement of the env-derived base.

NOTE: debugged and co-authored with Claude AI

### DIFF
--- a/internal/config/context.go
+++ b/internal/config/context.go
@@ -213,11 +213,17 @@ func fromOptions(ctx context.Context, opts *option.Options) context.Context {
 
 // ContextFromArgs returns a context.Context populated with any old-style
 // options or new-style arguments.
+//
+// The environment-derived context (see ContextFromEnv) is used as the base, so
+// variables like GHW_CHROOT continue to take effect when callers also pass
+// Modifiers or Options. Any value present in the supplied args wins over the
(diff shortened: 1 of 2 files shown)


jaypipes (approved):
@jimmykarily heyo! thank you very much for this submission (and the corresponding unit test). much appreciated! :)

PR #467, opened 2026-07-07
cpu: add regression test for non key:value lines in /proc/cpuinfo

europaul:
## What

Adds a targeted regression test (`TestQualcommCPUInfo`) for `/proc/cpuinfo` contents that
contain lines without a `:` separator, as seen on some Qualcomm ARM SoCs (e.g. QCM6125),
where the SoC name appears on a line of its own:

```
processor       : 0
...
CPU revision    : 2

Qualcomm Technologies, Inc QCM6125
```

## Why

Older ghw crashed on such lines. The underlying bug was already fixed upstream (#441, #443),
but there is no synthetic regression test locking that behavior in — the existing cpu tests
are all snapshot-fixture based. This test builds a minimal chroot with such a cpuinfo file
and asserts `cpu.New` succeeds and reports exactly one processor.

We (Zededa) hit this on real devices while using ghw for hardware inventory in
[lf-edge/eve](https://github.com/lf-edge/eve); this ports the test from our fork.

## Testing

`go test ./pkg/cpu/...` passes on Linux.

### DIFF
--- a/pkg/cpu/cpu_linux_test.go
+++ b/pkg/cpu/cpu_linux_test.go
@@ -219,3 +219,51 @@ func TestS390xCPU(t *testing.T) {
 		t.Fatal("Expected capabilities to be populated from s390x 'features' key")
 	}
 }
+
+func TestQualcommCPUInfo(t *testing.T) {
+	if _, ok := os.LookupEnv("GHW_TESTING_SKIP_CPU"); ok {
+		t.Skip("Skipping CPU tests.")
+	}
+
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
Looks great, thank you @europaul :)

jaypipes:
@europaul did you want to take this out of draft state?

europaul:
> @europaul did you want to take this out of draft state?

yes, sorry for the confusion :)

PR #470, opened 2026-07-08
Introduce pkg/tpm to detect and describe TPMs

europaul:
Adds a new `pkg/tpm` package that reports whether the host system has a Trusted Platform
Module and, when available, describes it:

- `TPMInfo.Present` is true when `/sys/class/tpm/tpm0` exists.
- `TPMInfo.Manufacturer`, `TPMInfo.SpecVersion` and `TPMInfo.FirmwareVersion` are parsed
  from the TPM 1.2 driver's `caps` file. TPM 2.0 devices expose no `caps` file, so the
  manufacturer falls back to the device `vendor` attribute and the spec version to the
  `tpm_version_major` attribute (kernel >= 5.5), yielding e.g. `2.0`.
- `SysClassTpm` is added to `linuxpath`, and the tpm class attributes are added to the
  snapshot clone content so detection also works against ghw snapshots.
- Wired into `HostInfo`, the `ghw` package aliases (`ghw.TPM()`, `ghw.TPMInfo`), a new
  `ghwc tpm` command (including the default human-format output), and the README.

## Testing

- New chroot-based unit tests cover TPM 1.2 `caps` parsing, the TPM 2.0
  vendor/`tpm_version_major` fallback, and the absent case.
- `go build ./... && go vet ./... && go test ./...` pass on Linux; cross-compiles for
  windows/darwin.
- Verified on a real machine with a TPM 2.0: `ghwc tpm` prints
  `tpm manufacturer= firmware_version= spec_version=2.0`.

### DIFF
--- a/pkg/linuxpath/path_linux.go
+++ b/pkg/linuxpath/path_linux.go
@@ -78,6 +78,7 @@ type Paths struct {
 	SysClassDRM            string
 	SysClassDMI            string
 	SysClassNet            string
+	SysClassTPM            string
 	SysClassWatchdog       string
 	SysFirmwareDeviceTree  string
 	RunUdevData            string
@@ -105,6 +106,7 @@ func New(ctx context.Context) *Paths {
 		SysClassDRM:            filepath.Join(chroot, roots.Sys, "class", "drm"),
 		SysClassDMI:            filepath.Join(chroot, roots.Sys, "class", "dmi"),
 		SysClassNet:            filepath.Join(chroot, roots.Sys, "class", "net"),
+		SysClassTPM:            filepath.Join(chroot, roots.Sys, "class", "tpm"),
 		SysClassWatchdog:       filepath.Join(chroot, roots.Sys, "class", "watchdog"),
 		SysFirmwareDeviceTree:  filepath.Join(chroot, roots.Sys, "firmware", "devicetree", "base"),
 		RunUdevData:            filepath.Join(chroot, roots.Run, "udev", "data"),
--- a/pkg/tpm/tpm.go
+++ b/pkg/tpm/tpm.go
@@ -1,0 +1,91 @@
+//
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
+package tpm
+
+import (
+	"fmt"
+
+	"github.com/jaypipes/ghw/internal/config"
+	"github.com/jaypipes/ghw/pkg/marshal"
+)
+
+// Device describes a single Trusted Platform Module (TPM) present on the host
+// system.
+type Device struct {
+	// ManufacturerName is the human-readable name of the TPM manufacturer,
+	// decoded from the four-character short name the TPM reports (e.g. "AMD",
+	// "IFX" for Infineon, "STM" for STMicroelectronics).
+	ManufacturerName string `json:"manufacturer_name"`
+	// ManufacturerVendorID is the 16-bit TCG Vendor ID assigned to the TPM
+	// manufacturer by the Trusted Computing Group. Note that this is a
+	// separate namespace from the PCI vendor ID. It may be empty when the
+	// host only exposes the manufacturer's short name.
+	// See: https://trustedcomputinggroup.org/resource/vendor-id-registry/
+	ManufacturerVendorID string `json:"manufacturer_vendor_id"`
+	// SpecVersion is the TCG TPM library (specification) version the device
+	// implements, i.e. the ISO/IEC 11889 version, e.g. "1.2" or "2.0".
+	SpecVersion string `json:"spec_version"`
+	// FirmwareVersion is the TPM firmware version, when the host exposes it.
+	FirmwareVersion string `json:"firmware_version"`
+}
+
+// String returns a short string describing the TPM device.
+func (d *Device) String() string {
+	return fmt.Sprintf(
+		"manufacturer_name=%s manufacturer_vendor_id=%s firmware_version=%s spec_version=%s",
+		d.ManufacturerName, d.ManufacturerVendorID, d.FirmwareVersion, d.SpecVersion,
+	)
+}
+
+// Info describes the Trusted Platform Module(s) on the host system.
+type Info struct {
+	// Devices are the TPM devices detected under /sys/class/tpm. A host can,
+	// in principle, expose more than one TPM device.
+	Devices []*Device `json:"devices"`
+}
+
+// String returns a short string with summary information about the TPM(s) on
+// the host system.
+func (i *Info) String() string {
+	switch len(i.Devices) {
+	case 0:
+		return "tpm (no devices)"
+	case 1:
+		return "tpm " + i.Devices[0].String()
+	default:
+		return fmt.Sprintf("tpm (%d devices)", len(i.Devices))
+	}
+}
+
+// New returns a pointer to an Info struct that contains information about
+// the TPM(s) on the host system.
+func New(args ...any) (*Info, error) {
+	ctx := config.ContextFromArgs(args...)
+	info := &Info{}
+	if err := info.load(ctx); err != nil {
+		return nil, err
+	}
+	return info, nil
+}
+
+// simple private struct used to encapsulate TPM information in a top-level
+// "tpm" YAML/JSON map/object key
+type tpmPrinter struct {
+	Info *Info `json:"tpm"`
+}
+
+// YAMLString returns a string with the TPM information formatted as YAML
+// under a top-level "tpm:" key
+func (i *Info) YAMLString() string {
+	return marshal.SafeYAML(tpmPrinter{i})
+}
+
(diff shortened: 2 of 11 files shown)


pkg/tpm/tpm.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+
+	"github.com/jaypipes/ghw/internal/config"
+	"github.com/jaypipes/ghw/pkg/marshal"
+)
+
+// Info describes the Trusted Platform Module (TPM) on the host system.
+type Info struct {
+	Present         bool   `json:"present"`
```
jaypipes: Instead of a Present boolean field, let's make a `Device` struct and populate a `Info.Devices` field of type `[]Device` similar to [how the pci module works](https://github.com/jaypipes/ghw/blob/main/pkg/pci/pci.go#L22). This way we keep the interface between the two modules similar (plus, technically a system can have >1 TPM device, so we make this interface more future-proof)

pkg/tpm/tpm.go
```diff
@@ -0,0 +1,63 @@
...
+	Manufacturer    string `json:"manufacturer"`
```
jaypipes: Let's name this field `ManufacturerName` and add a separate `ManufacturerVendorID` field to store the [16-bit TCG Vendor ID](https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-Vendor-ID-Registry-Family-1.2-and-2.0-Version-1.06-Revision-0.96_pub.pdf). I wish that TCG just used the PCI vendor ID but alas, they did not :)

pkg/tpm/tpm.go
```diff
@@ -0,0 +1,63 @@
...
+	FirmwareVersion string `json:"firmware_version"`
```
jaypipes, line 33: This is only relevant for fTPMs, not dTPMs, sTPMs, iTPMs or vTPMs. I think we should add a `Type` field to the `Device` struct that indicates whether the TPM is a `Firmware`, `Discrete`, `Software`, `Integrated` or `Virtual` TPM type. Thoughts?

On Linux, right now there's not a great way to determine the type other than to just say everything is a Firmware TPM unless it's Infineon, which is a discrete TPM. On Windows, there's better ways to determine the TPM type...

pkg/tpm/tpm.go
```diff
@@ -0,0 +1,63 @@
...
+	SpecVersion     string `json:"spec_version"`
```
jaypipes: Please add a little docstring above these fields explaining what they contain. In this case, it is the TCG TPM Library version (ISO/IEC 11889).

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+	if _, err := os.Stat(tpmPath); err != nil {
+		i.Present = false
+		return nil
+	}
+	i.Present = true
+
+	// TPM 1.2 drivers expose a caps file with manufacturer, spec and
+	// firmware version information.
+	if file, err := os.Open(filepath.Join(tpmPath, "caps")); err == nil {
+		defer file.Close()
```
jaypipes: To get rid of the linter failure, use [`util.SafeClose()`](https://github.com/jaypipes/ghw/blob/main/pkg/util/util.go#L27)

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/jaypipes/ghw/pkg/linuxpath"
+	"github.com/jaypipes/ghw/pkg/util"
+)
+
+func (i *Info) load(ctx context.Context) error {
+	paths := linuxpath.New(ctx)
+	tpmPath := filepath.Join(paths.SysClassTpm, "tpm0")
```
jaypipes: As mentioned earlier, there can be more than one TPM device on a system, so here we will want to loop over any devices not just check for "tpm0".

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+			parts := strings.SplitN(line, ":", 2)
+			if len(parts) != 2 {
+				continue
+			}
+			key := strings.TrimSpace(parts[0])
+			val := strings.TrimSpace(parts[1])
+
+			switch key {
+			case "Manufacturer":
+				i.Manufacturer = val
```
jaypipes: This will be the hex-encoded string value of the Manufacturer's string name. For example, `0x414D4400` for `"AMD"`. I think we should convert the hex-encoded string value to ASCII here.

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+				i.SpecVersion = val
+			case "Firmware version":
+				i.FirmwareVersion = val
+			}
+		}
+	}
+
+	if i.Manufacturer == "" {
+		if b, err := os.ReadFile(filepath.Join(tpmPath, "device", "vendor")); err == nil {
+			i.Manufacturer = strings.TrimSpace(string(b))
```
jaypipes: For TPM 2.0 devices, this *may* be the 16-bit TCG Vendor ID, hex-encoded string value. For example, `0x1414` for Microsoft or `0x15D1` for Infineon. Or it may be a 32-bit hex-encoded string value, for example `0x414D4400` for `"AMD"` ([the underlying type is a 32-bit integer in the Linux TPM module header](https://github.com/torvalds/linux/blob/0e35b9b6ec0ffcc5e23cbdec09f5c622ad532b53/drivers/char/tpm/tpm.h#L132)).

To deal with this mess, it's probably best to do a check here (and potentially above on line 46) to see whether the length of the hex-encoded string is 8 or 16. If it's 8, then this is the 16-bit TCG Vendor ID. If it's 16, this is the 32-bit Manufacturer short name. Otherwise, it's a long-form Manufacturer string.

There has unfortunately not been standards applied to sysfs in this area :(

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+func (i *Info) load(ctx context.Context) error {
+	paths := linuxpath.New(ctx)
+	tpmPath := filepath.Join(paths.SysClassTpm, "tpm0")
+	if _, err := os.Stat(tpmPath); err != nil {
+		i.Present = false
+		return nil
+	}
+	i.Present = true
+
+	// TPM 1.2 drivers expose a caps file with manufacturer, spec and
```
jaypipes: TPM 2.0 drivers may also expose this caps file... it's unfortunately not been standardized. But, you catch this below in the block that looks for no SpecVersion being set, so we're good (just update the code comment above)

pkg/tpm/tpm_linux.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if i.Manufacturer == "" {
+		if b, err := os.ReadFile(filepath.Join(tpmPath, "device", "vendor")); err == nil {
+			i.Manufacturer = strings.TrimSpace(string(b))
+		}
+	}
+
+	// TPM 2.0 devices have no caps file; the major spec version is exposed
+	// separately (kernel >= 5.5).
```
jaypipes: Technically, TPM 2.0 devices *can* have a caps file exposed via sysfs. But the SpecVersion field in that file will not be set.

jaypipes (changes requested):

pkg/tpm/tpm.go
europaul: I couldn't find a reliable way to differentiate between those types... I think if vTPM is done properly then it will be indistinguishable from dTPM - tried that on our EVE in QEMU with swtpm.

We could introduce the `Type` field and set it to `Unknown` for Linux in most cases and otherwise try to determine the TPM type from the driver that it uses, but I'm not sure if it helps a lot.

Maybe leave it for later when somebody implements the Windows part?

jaypipes: @europaul yeah, totally cool to leave the Type for later :)

europaul:
@jaypipes I think I addressed the rest of the comments. Please have another look.

pkg/linuxpath/path_linux.go
```diff
@@ -78,6 +78,7 @@ type Paths struct {
 	SysClassDRM            string
 	SysClassDMI            string
 	SysClassNet            string
+	SysClassTpm            string
```
jaypipes: @europaul sorry, just noticed this... please make this `SysClassTPM`.

See here about recommended Go [variable names involving initialisms](https://google.github.io/styleguide/go/decisions.html#initialisms).

jaypipes (changes requested):
One teeny nit to address, then all good! Thanks @europaul :)

pkg/linuxpath/path_linux.go
europaul: done :+1: I also squashed the commit to one now so everything is clean

jaypipes (approved):
Brilliant, thank you @europaul :)

PR #471, opened 2026-07-09
linux: fall back to DeviceTree when DMI is unavailable

rpardini:
DMI/SMBIOS is absent on most DeviceTree-based systems (ARM/RISC-V boards), leaving chassis, product, baseboard, bios and cpu info all unknown. When DMI is unavailable but the DeviceTree is present, populate identity from /sys/firmware/devicetree/base instead:

  - chassis/product/baseboard: model, vendor (compatible prefix), serial-number
  - chassis type: map DeviceTree chassis-type to the SMBIOS type code
  - bios: `chosen/u-boot,version` reported as U-Boot firmware
  - cpu: SoC compatible (last entry) as processor model when cpuinfo has none

DMI remains preferred when present. All DeviceTree property knowledge is encapsulated in the new `linuxdt` package via semantic accessors.

### DIFF
--- a/pkg/linuxdt/dt_linux.go
+++ b/pkg/linuxdt/dt_linux.go
@@ -1,0 +1,149 @@
+// Use and distribution licensed under the Apache license version 2.
+//
+// See the COPYING file in the root project directory for full text.
+//
+
+// Package linuxdt reads identity information from the Linux DeviceTree, exposed
+// by the kernel under /sys/firmware/devicetree/base. It is the DeviceTree
+// counterpart to the linuxdmi package and is used as a fallback on systems (most
+// ARM/RISC-V boards) that have no DMI/SMBIOS tables.
+//
+// All knowledge of DeviceTree property names and layout lives here: callers ask
+// for a piece of identity (Model, Vendor, SerialNumber, ...) and never deal with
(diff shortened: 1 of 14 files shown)


pkg/linuxdt/dt_linux.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// propertyList reads a NUL-separated string list property (for example
+// "compatible") from the DeviceTree base node. It returns nil if the property is
+// missing or cannot be read.
+func propertyList(ctx context.Context, name string) []string {
+	paths := linuxpath.New(ctx)
+	path := filepath.Join(paths.SysFirmwareDeviceTree, name)
+
+	log.Debug(ctx, "reading from %q", path)
```
jaypipes, line 134: Probably best to remove this :) I need to remove it from the similar place in linuxdmi....

jaypipes (approved):
@rpardini this is excellent. Thank you so much for this contribution!

There's room to clean up some of the repeated testing utilities and I'd definitely want to get some ghw snapshots added to the `testdata/` directory for these systems you are working with. But that can always be done in a followup PR. :)

Welcome to the `ghw` contributor club!

PR #474, opened 2026-09-18
fix(block/darwin): detect NVMe controller on Apple Silicon

mdevils:
## Summary

`storageControllerFromPlist` only matched `IONVMeController` in the device tree path to detect NVMe. On Apple Silicon Macs the internal SSD is attached through the SoC's own storage fabric instead, and the device tree node name varies by chip generation (`AppleANS2Controller` on T2 Macs, `AppleANS3NVMeController` on M-series Macs, etc.), so it never matched and every internal disk was misclassified as `StorageControllerSCSI`.

Fix: also match `BusProtocol == "Apple Fabric"`, which macOS reports consistently for this bus regardless of the exact controller node name.

Verified against a real M-series Mac: `Disk.StorageController` now correctly reports `NVMe` for the internal SSD instead of `SCSI`.

## Test plan

- [x] Added `pkg/block/block_darwin_test.go` covering: standard `IONVMeController` path, Apple Silicon `AppleANS3NVMeController` + `Apple Fabric`, T2 `AppleANS2Controller` (no "NVMe" substring in path) + `Apple Fabric`, and a non-Apple-Fabric SATA/USB disk staying `SCSI`.
- [x] `go build ./...` and `go vet ./pkg/block/...`
- [x] Manually verified with `ghw.Block()` against real hardware (M-series MacBook Pro internal SSD)


jaypipes (approved):
Thanks very much for this contribution @mdevils! Much appreciated :) Code and tests lgtm.

